PR #40, opened 2018-05-10
Transfer label indicates the direction now. Localization added.

MaaKut:


zyuhel:
Please rename constants from `amount_sent` to `sent_label`, and `amount_received` to `received_label`

PR #41, opened 2018-05-11
New chats will now start empty

MaaKut:


zyuhel:
If you come back from newly opened chat to chat list, you will get error with data showed. It shouldn't render in that case.

screen shot 2018-05-19 at 11 29 21

PR #42, opened 2018-05-22
Feature/ethereum base support

MaaKut:


zyuhel:
Imo Vscode settings should be added to .gitignore and not uploaded to code repository

PR #44, opened 2018-05-23
Health check

hdkeeper:


src/lib/healthCheck.js
```diff
@@ -0,0 +1,89 @@
...
+    if (this.intId) {
+      clearInterval(this.intId)
+      this.intId = null
+    }
+  }
+
+  // Probe given urls
+  probe () {
+    // Wait for all probes
+    Promise.all(this.urls.map(o => new Promise((resolve, reject) => {
```
zyuhel: Maybe better replace 
````o => new Promise((resolve, reject) => {
      this.requester.get(makeUrl(o))
        .then(() => resolve(true))
        .catch(() => resolve(false))
   })
````
with 
```` o => this.requester.get(makeUrl(o)).then(() => true, () => false) ````

hdkeeper: Well, without call to resolve() it will not work as designed.
I have just tested this thing.

hdkeeper:
All fixes are here, so you can merge it.

PR #50, opened 2018-06-01
add qr code login

chovy:
return passphrase from store


src/components/Login.vue
```diff
@@ -24,9 +24,78 @@
...
               <md-layout md-align="center" md-gutter="16">
                   <md-button class="md-raised md-short" v-on:click="logme">{{ $t('login.login_button') }}</md-button>
-
               </md-layout>
           </md-layout>
+          <md-layout md-flex="66" md-flex-xsmall="80" md-align="center">
+            <md-button classs="md-short"
+                      @click.prevent="scanQRCode">
+              <!-- Generator: Adobe Illustrator 19.0.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+              <svg version="1.1" id="Capa_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
```
MaaKut: Let's move this SVG to a separate file

chovy: I will create an <icon> component that allows us to use svg icons in the dom.

PR #70, opened 2018-06-15
fixups on qr code scan

chovy:
- show tooltip on button
- display invalid passphrase when qr scan fails
- redirect back to homepage


zyuhel:
english text tooltip variable should be added to all languages, or languages should be setup to show english if no constant is available.

chovy:
Should I just use google translate to get the other languages?

PR #73, opened 2018-06-19
Added welcome messages

onemorejsdeveloper:
According to spec were added generated welcome messages


zyuhel:
Addresses in welcome messages (for example U15677078342684640219 )  should be links opening new chat with that user.

PR #75, opened 2018-06-22
Fix bug with encoded pass phrase

onemorejsdeveloper:
Fixed app crash, if pass phrase was not encoded

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "adamant-im",
-  "version": "1.1.1",
+  "version": "1.1.2",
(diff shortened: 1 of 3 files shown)


zyuhel:
May be it will be faster to check if passphrase is twelve words, instead of using base64 check regex?

zyuhel:
It will be around 30% faster to check last symbol instead of such complex regexp - http://jsben.ch/6K19u 
But this function generally run only once on each page load. So maybe it should be optimized prematurely.  
But 0.1s delay before user can interact with page is pretty bad thing imo.

PR #81, opened 2018-06-29
Delegates and votes

sergushakov:
Votes screen completed


zyuhel:
Delegate registration is removed, so supporting code also should be removed as well

src/store/index.js
```diff
@@ -153,6 +156,9 @@ const store = {
       state.showPanel = false
       state.showBottom = true
       state.transactions = {}
+      state.delegates = {}
+      state.delegates = {}
```
zyuhel: unnecessary variable duplication

src/lib/adamantServerApi.js
```diff
@@ -519,6 +520,216 @@ function install (Vue) {
...
     })
   }
+
+  Vue.prototype.getDelegates = function (limit = constants.Delegates.ACTIVE_DELEGATES, offset = 0, votes = []) {
+    this.$store.commit('ajax_start')
+    this.$http.get(this.getAddressString() + `/api/delegates?limit=${limit}&offset=${offset}`).then(response => {
+      if (response.body.success) {
+        for (let i in response.body.delegates) {
+          let delegate = response.body.delegates[i]
+          let voted = votes.includes(delegate.address)
```
MaaKut: `let` should be used to the mutable values. If variable value is not going to change, please, use `const` instead.

src/lib/adamantServerApi.js
```diff
@@ -519,6 +520,216 @@ function install (Vue) {
...
+    }, () => {
+      this.$store.commit('ajax_end_with_error')
+    })
+  }
+  Vue.prototype.voteForDelegates = function (votes) {
+    let keys = this.getKeypair()
+    let transaction = {
+      type: constants.Transactions.VOTE,
+      asset: {votes: votes},
+      timestamp: adamant.epochTime(),
```
MaaKut: Client local time may not be in sync with the Adamant node time, which may cause the transaction to fail. `adamant-api.js` has a workaround for this.

src/lib/adamantServerApi.js
```diff
@@ -519,6 +520,216 @@ function install (Vue) {
...
+        })
+        this.$http.get(this.getAddressString() + '/api/blocks?orderBy=height:desc&limit=100').then(response => {
+          if (response.body.success) {
+            let lastBlock = response.body.blocks[0]
+            let blocks = response.body.blocks.filter(x => x.generatorPublicKey === delegate.publicKey)
+            let time = Date.now()
+            let status = { updatedAt: time }
+            if (blocks.length > 0) {
+              status.lastBlock = blocks[0]
+              status.blockAt = new Date((((Date.UTC(2017, 8, 2, 17, 0, 0, 0) / 1000) + status.lastBlock.timestamp) * 1000))
```
MaaKut: Please, use relevant constants from `constants.js` instead of `Date.UTC(2017, 8, 2, 17, 0, 0, 0)`

src/lib/adamantServerApi.js
```diff
@@ -519,6 +521,213 @@ function install (Vue) {
...
+      } else {
+        this.$store.commit('ajax_end_with_error')
+      }
+    }, () => {
+      this.$store.commit('ajax_end_with_error')
+    })
+  }
+  Vue.prototype.voteForDelegates = function (votes) {
+    let keys = this.getKeypair()
+    let transaction = adamantAPI.newTransaction(constants.Transactions.VOTE)
```
MaaKut: Why? Move the `api/accounts/delegates` call to the `adamant-api.js` entirely; there's no need to export `newTransaction` and thus break the incapsulation.

src/lib/adamantServerApi.js
```diff
@@ -519,6 +521,213 @@ function install (Vue) {
...
+    const round = height => {
+      if (isNaN(height)) {
+        return 0
+      } else {
+        return Math.floor(height / 101) + (height % 101 > 0 ? 1 : 0)
+      }
+    }
+
+    this.$store.commit('ajax_start')
+    this.$http.get(this.getAddressString() + '/api/delegates/getNextForgers?limit=101').then(response => {
```
MaaKut: Would be better to use `Delegates.ACTIVE_DELEGATES` constant insread of magic number.

PR #123, opened 2018-08-20
Fix reset state with transactions list

onemorejsdeveloper:


onemorejsdeveloper:
I've fixed missing public ETH address

PR #128, opened 2018-08-29
Login via password

onemorejsdeveloper:


src/components/Login.vue
```diff
@@ -15,31 +15,23 @@
           <p class="site-description">{{ $t('login.subheader') }}</p>
       </div>
       <md-layout md-align="center" md-gutter="16">
-      <md-layout md-flex="66" md-flex-xsmall="80">
-          <md-input-container class="password_input">
+          <LoginWithUserPassword v-if="userPassword" ref="login_with_user_password"></LoginWithUserPassword>
+          <md-layout v-else md-flex="66" md-flex-xsmall="80">
+            <md-input-container class="password_input">
               <label style="text-align: center;width: 100%;">{{ $t('login.password_label') }}</label>
               <md-input v-model="passPhrase" type="password" autocomplete="new-password" @keyup.native="kp($event)"></md-input>
```
zyuhel: Why do we need such autocomplete value?

onemorejsdeveloper: Well, it is previuos implementation. If need to remove it, so let's do it

onemorejsdeveloper:
So, I have fixed PWA according notes of @adamant-al. Please, could mind checking it?

adamant-al:
-2) Keep login via "Enter" key.
Snackbar fires even if password correct.

-3) Check button colors.

-9) @zyuhel check and update here https://docs.google.com/document/d/15ONBizk_UqdT7GiqL5bMfr8g5R32VXPJZIckNE19crk/edit#bookmark=id.f9nn16yzbw2s

-10) Logout button (bottom  bar) should work as "Remove password" — clear everything totally.

adamant-al:
-11) Set focus on pass-edit after load on both Login and Password screens

zyuhel:
Password input field should be password, not text. it may be switchable, but input must be hidden by default. 
While encrypting on Set button, we should change button to loader or somehow show that process is going on.

adamant-al:
Password set, open in new tab https://loginviapassword-branch-adamant-im.surge.sh/ — asking for passphrase. Expected: ask for password.

adamant-al:
-3.
indexedDb.js:133 Uncaught (in promise) TypeError: Cannot read property 'toString' of null
    at B (indexedDb.js:133)
Once again:
Test with opened debug console!

-4.
I still see all the info in Local/Session storage.

-5.
Password set, open in new tab https://loginviapassword-branch-adamant-im.surge.sh/ — asking for password.
Open console, I see all the data decripted in IndexedDB before I enter password.
Kindergarden?

RealGoodProgrammer:
Run linter and correct errors

adamant-al:
1-
When click Logout, Session Storage are not cleared.

2-
When set password, Session Storage are not cleared.

3-
Store chatList encrypted.

Elena-Andreeva:
1. If user saves passphrase in Chrome, it automatically appears in "password" window. 
2. You can insert Agreement text (now it leads to our website).
Text have already written: 

When you login using password, you are aware of possible reduction of security level.  
You also confirm that you’ve read this article and learned about ADAMANT Messenger Terms of Service (https://adamant.im/terms/). Remember: safety of an account is your own responsibility.

zyuhel:
On password login, Address Book is not restored.

Elena-Andreeva:
On password login, Address Book was removed. When I accesed by passphrrase, contacts' names wasn't appear.

PR #135, opened 2018-09-12
Bug/fix f5 and styles

RealGoodProgrammer:


src/App.vue
```diff
@@ -207,7 +207,7 @@ footer {
     min-width: 70%;
 }
 .md-toolbar .md-title .md-input-container:after {
-    background: none ;
+    background: none !important;
```
zyuhel: can we make more specified selector, add additional class, instead of using important.

src/App.vue
```diff
@@ -222,6 +222,9 @@ footer {
     top: 13px;
     font-size: 18px;
 }
+.md-theme-grey.md-input-container.md-input-focused label {
+  color: rgba(0, 0, 0, 0.54) !important;
```
zyuhel: can we make more specified selector, add additional class, instead of using important.

src/App.vue
```diff
@@ -269,7 +272,12 @@ footer {
 .chats .md-list {
     max-width: 90%;
 }
-
+a:not(.md-button) {
+  color: #2e7eed!important;
```
zyuhel: can we make more specified selector, add additional class, instead of using important.

src/App.vue
```diff
@@ -269,7 +272,12 @@ footer {
...
+}
+.login .create_link {
+  color: #4a4a4a!important;
```
zyuhel: can we make more specified selector, add additional class, instead of using important.

src/App.vue
```diff
@@ -354,6 +362,12 @@ header span {
 .md-dialog {
   box-shadow: initial;
 }
+.md-dialog .md-primary {
+  background: #ffffff;
+}
+.md-dialog-confirm > .md-dialog > .md-dialog-actions > button{
+  color: #000 !important;
```
zyuhel: can we make more specified selector, add additional class, instead of using important.

PR #148, opened 2018-09-28
Modern icons

RealGoodProgrammer:
New PWA icons


zyuhel:
Safari pinned tab is not contrast with such complex icon, we should either change icon, or change colour

PR #162, opened 2018-11-04
Fixed unexpected br tag in chat

onemorejsdeveloper:


Elena-Andreeva:
Unexpected tag <p> in chat when you send message after wrong sending
http://dl4.joxi.net/drive/2018/11/07/0021/3536/1424848/48/095c948b76.jpg

src/store/index.js
```diff
@@ -187,19 +186,19 @@ const store = {
...
-      const message = currentChat.messages[payload]
-      let messageText = message.message.replace(/<\/?p>/g, '')
+      let message = currentChat.messages[payload]
+      let messageText = message.message
+      messageText = messageText.replace(/<p>|<\/?p>/g, '')
+      messageText = messageText.replace(/<br>/g, '\n')
       payload = {
         recipientId: partner,
-        message: messageText,
+        message: message.message,
```
zyuhel: why here is message.message? it should be either messageText or code above doesn't make sense

PR #167, opened 2018-11-19
Feature/eth tx list

MaaKut:


Elena-Andreeva:
1. There is no chronological order in the transaction list: http://dl3.joxi.net/drive/2018/11/28/0021/3536/1424848/48/4d09057f3e.jpg
2. Old bug: а new chat doesn't starts.

PR #169, opened 2018-11-21
Added promotion AppBar for iOS

RealGoodProgrammer:
Яблочники, тестируйте)


RealBonus:
Smart banners supported only in iOS Safari.

PR #189, opened 2018-12-27
Fix undesired links rendering

zyuhel:


zyuhel:
We shouldn't render as link all two word combinations separated with dots, like not.link

PR #190, opened 2018-12-27
Adamantpwa 103

onemorejsdeveloper:
Fix problem when sometimes chats were marked as unread after login with password.


src/views/Chats.vue
```diff
@@ -64,9 +64,6 @@ export default {
   name: 'chats',
   components: { NewChat, ChatEntry },
   methods: {
-    load () {
```
onemorejsdeveloper: Removed as unused code

PR #209, opened 2019-01-28
feat(Transactions.vue): display text if no transactions

bludnic:


src/views/Transactions.vue
```diff
@@ -1,34 +1,39 @@
...
+          </md-list-item>
+        </md-list>
 
-          <md-divider class="md-inset"></md-divider>
-        </md-list-item>
-        <md-list-item class="transaction_list__loader" v-if="isLoading">
-          <inline-spinner />
-        </md-list-item>
-      </md-list>
+        <h3 v-else class="md-headline">{{ $t('transaction.no_transactions') }}</h3>
```
MaaKut: In an `if ... else ...` expression the shorter branch should go first for better readability.

src/views/Transactions.vue
```diff
@@ -1,34 +1,39 @@
...
-        <md-list-item v-for="(transaction) in transactions" :key="transaction.id" style="cursor:pointer">
-          <md-avatar>
-            <md-icon>{{ transaction.direction === 'from' ? 'flight_takeoff' : 'flight_land' }}</md-icon>
-          </md-avatar>
 
-          <div class="md-list-text-container" v-on:click="goToTransaction(transaction.id)">
-            <div>
-              {{ displayName(transaction) }}
-              <span class="partner_display_name">{{ formatPartnerAddress(transaction) }}</span>
+      <template v-if="isFulfilled">
```
MaaKut: Cannot we use `areTransactionsLoading` instead? Also, as far as I understand, this component will display nothing while `isFulfilled == false` - is this correct? I believe, some kind of a spinner should be displayed.

PR #211, opened 2019-01-30
fix(eth-base): convert `gasPrice` from hex when `calculateFee`

bludnic:


src/store/modules/eth-base/eth-base-actions.js
```diff
@@ -198,10 +198,10 @@ export default function createActions (config) {
       const gasPrice = transaction.gasPrice
 
       const supplier = () => api.eth.getTransactionReceipt.request(payload.hash, (err, tx) => {
-        if (!err && tx && checkBlockCount(tx, context.rootState)) {
+        if (!err && tx && checkBlockCount(tx, context.rootState) && gasPrice) {
```
MaaKut: Do we really need this check here?

PR #213, opened 2019-01-30
fix(Transactions.vue): display in-chat transfers also

bludnic:
fix(lcStore): do not store `areChatsLoading` inside SS, in order not to
block `actions.loadChats` execution

Transfer.vue: display `comments` field when sending ADM from chat
adamant-api: fetch transactions + in-chat transactions
package.json: add lodash

### DIFF
--- a/src/lib/lsStore.js
+++ b/src/lib/lsStore.js
@@ -1,4 +1,5 @@
 import merge from 'deepmerge'
+import cloneDeep from 'lodash/cloneDeep'
 import {
   decryptData,
   encryptData,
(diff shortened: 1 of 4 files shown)


src/lib/lsStore.js
```diff
@@ -181,7 +183,9 @@ export default function storeData () {
         let storeNow = false
         if (mutation.type === 'change_storage_method') {
           if (!mutation.payload) {
-            useStorage = sessionStorage.setItem('adm-persist', JSON.stringify(state))
+            const clonedState = cloneDeep(state)
+            delete clonedState.areChatsLoading
+            useStorage = sessionStorage.setItem('adm-persist', JSON.stringify(clonedState))
```
MaaKut: Did you just add a whole `lodash` as a dependency just to remove a property from the persisted state? If yes, please, undo this. Instead just use:
```sessionStorage.setItem('adm-persist', JSON.stringify(state, (k, v) => (k === 'areChatsLoading' ? undefined : v)))```

PR #221, opened 2019-02-09
App optimization

bludnic:
- сache `state.{publicKey, privateKey}` instead of calling `unlock()` after refresh
- refactor `partners` module and rename to `contacts` & add unit tests
- LBP: save to indexedDb `publicKey` & `privateKey` also
- F5: add `AppInit.js` to restore application state
- remove `Base64.{encode/decode}` (slow synchronous code)


src/lib/adamant-api.js
```diff
@@ -13,13 +13,22 @@ Queue.configure(Promise)
...
 
-/** @type {{privateKey: Buffer, publicKey: Buffer}} */
-let myKeypair = { }
-let myAddress = null
-
 /** Lists cryptos for which addresses are currently being stored to the KVS */
 const pendingAddresses = { }
 
+/** Vuex selectors **/
+const user = {
```
MaaKut: What does store do here? `adamant-api` is a lower-level layer and should know nothing about the store.

src/lib/idb/state.js
```diff
@@ -5,7 +5,7 @@ import Chats from './stores/Chats'
 import Security from './stores/Security'
 
 /** Modules that will be stored in IDB **/
-const modules = ['adm', 'eth', 'doge', 'bnb', 'bz', 'partners', 'delegates']
+const modules = ['adm', 'eth', 'doge', 'bnb', 'bz', 'contacts', 'delegates']
```
MaaKut: Coin-modules (`eth`, `doge`, etc.) should be added dynamically (see `constants/Cryptos`).

bludnic: What do you suggest?
`const modules = ['partners', 'delegates'].concat(Object.values(Cryptos).map(crypto => crypto.toLowerCase()))`

PR #238, opened 2019-03-01
SendFundsForm: show spinner on `Confirm` button

bludnic:


zyuhel:
Maybe we should add v-btn--flat to confirm button, so that it looks more similar to cancel button and master branch?

PR #240, opened 2019-03-02
style Transactions like in master-branch

bludnic:


zyuhel:
There is no start chat icon in transaction list, if there was no chat with that user. On master it is shown with chat_bubble_outline material design icon.

PR #243, opened 2019-03-10
Add animation to language selector and icons

zyuhel:
Little animation never hurt anyone

### DIFF
--- a/src/views/Login.vue
+++ b/src/views/Login.vue
@@ -199,14 +199,17 @@ export default {
     font-family: 'Exo 2'
     font-weight: 100
     font-size: 18px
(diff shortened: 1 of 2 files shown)


src/views/Login.vue
```diff
@@ -199,14 +199,17 @@ export default {
     font-family: 'Exo 2'
     font-weight: 100
     font-size: 18px
-
+.login__icon, .v-icon--link
```
bludnic: Do you mean `.login-page__icon`?

zyuhel: yes, you are right, fixed

PR #251, opened 2019-03-26
Feature/electron build all

zyuhel:
Add command for building all platforms, will be usable for travis integration


RealGoodProgrammer:
It will work only on OS X platform.

PR #252, opened 2019-03-26
Reduce bundle size

bludnic:
- tree-shaking `bitcore-mnemonic/lib/words/*` except english.js
- replace `moment` with `dayjs`
- replace `instascan` & `vue-qrcode-reader` with `@zxing/library`


zyuhel:
We shouldn't show selector if there is only one option. Also it is better to switch between sources using icon over video, instead of select drop down.

zyuhel:
bip39 have released new version, it can be treeshaked using webpack's ignore plugin, i think we should update to it.

zyuhel:
Do we need empty.loader.js now?

PR #257, opened 2019-04-12
Direct transfer

bludnic:


src/components/SendFundsForm.vue
```diff
@@ -313,9 +306,8 @@ export default {
       if (this.currency === Cryptos.ADM) {
         let promise
         // 1. sendMessage if come from chat
```
zyuhel: We should send message if there is a comment, not when in-chat

zyuhel:
When sending in-chat transfer without a comment, it doesn't appear in chat, although it is sent.

PR #260, opened 2019-04-19
Save current state of WalletTab in LS

bludnic:


zyuhel:
We need to store state for the use-case of going to history view and back.
Also i don't think that selected tab should be stored between sessions. But i think that behaviour should be discussed.

PR #269, opened 2019-05-19
Fix passphrase validation on Login View

bludnic:


src/components/LoginForm.vue
```diff
@@ -44,6 +44,10 @@
 </template>
 
 <script>
+function validatePassphrase (passphrase) {
+  return passphrase.toLowerCase().trim().split(' ').length === 12
```
zyuhel: Why do we need toLowerCase here?

zyuhel: Also maybe instead of counting spaces we should use bip39.validateMnemonic(mnemonic)?

PR #302, opened 2019-06-27
AIP 2: URI Format for ADAMANT

bludnic:


bludnic:
@adamant-al It's just a test that check if the parameters are parsed correctly. The current implementation handles only `label` parameter, as described in the specification.

PR #337, opened 2019-07-31
Scroll chat to the bottom when window activated by desktop notification

graycraft:


src/lib/notifications.js
```diff
@@ -70,6 +73,7 @@ class PushNotification extends Notification {
                   closeOnClick: true,
                   icon: 'img/icons/android-chrome-192x192.png',
                   notifyClick: () => {
+                    this.increaseCounter()
```
zyuhel: We increase counter as workaround to easily detect need of scrolling to bottom in currently opened component

PR #344, opened 2019-08-06
Update en.jsor

OlgaDrabenya:


adamant-al:
Need PR to dev, not in master

PR #436, opened 2022-09-08
Fix/unit tests

bludnic:
- [x] Updated old snapshots
- [x] Fixed failed tests of several components
- [x] Fix `SyntaxError: Cannot use import statement outside a module` in some test files
- [x] Fix dozens of tests as a side effect after fixing previous issue

https://trello.com/c/2yGMnM53


adamant-al (approved):

src/lib/__tests__/uri.test.js
```diff
@@ -53,21 +66,32 @@ describe('uri', () => {
...
+
+    afterAll(() => {
+      delete window.location;
+    })
+
     it('without contactName', () => {
-      expect(generateURI('U123456')).toBe('adm:U123456')
-      expect(generateURI('U123456', '')).toBe('adm:U123456')
-      expect(generateURI('U123456', undefined)).toBe('adm:U123456')
+      expect(generateURI(Cryptos.ADM, 'U123456')).toBe('https://msg.adamant.im?address=U123456')
```
martiliones, line 87: I think we need to create a constant for base url `https://msg.adamant.im`

src/components/__tests__/__snapshots__/AppNavigation.test.js.snap
```diff
@@ -78,7 +82,9 @@ exports[`AppNavigation.vue renders the correct markup 1`] = `
           <v-icon-stub
             size="20"
           >
-            mdi-settings
+
```
martiliones, line 85: why are there new lines?

src/components/__tests__/AppNavigation.test.js
```diff
@@ -119,24 +119,54 @@ describe('AppNavigation.vue', () => {
     expect(wrapper.vm.getCurrentPageIndex()).toBe(2)
   })
 
-  it('should display number of new messages', () => {
+  it('badge should be hidden when there are no messages', () => {
```
martiliones, line 122: "it badge should be..." is incorrect. Test is describing 'AppNavigation.vue', so you should write tests on its behalf:

```js
it('should hide the badge when there are no messages', /* ... */)
```
and it should be like that everywhere

src/components/__tests__/SendFundsForm.test.js
```diff
@@ -347,13 +385,18 @@ describe('SendFundsForm', () => {
       )
     })
 
-    it('confirm message with `recipientName`', () => {
+    it('confirm message with `recipientName`', async () => {
```
martiliones, line 388: should confirm..

src/components/__tests__/ChatStartDialog.test.js
```diff
@@ -30,32 +30,51 @@ function createChat (context, { partnerId }) {
...
+    passphrase: ''
+  })
   const snackbar = mockupSnackbar()
   const chat = {
+    getters: {
+      isAdamantChat: () => () => false,
+    },
     actions: {
-      createChat: createChat
+      createChat: createChat,
```
martiliones, line 44: use ES6 features
`createChat: createChat` -> `createChat`

martiliones (changes requested):

src/components/__tests__/AppNavigation.test.js
bludnic: The "badge" is part of the AppNavigation component, so it should be tested within this component.

src/components/__tests__/ChatStartDialog.test.js
bludnic: We will fix such cases later with `prettier` across the entire codebase. That part wasn't touched so I suggest leave it as is.

src/lib/__tests__/uri.test.js
bludnic: Normally the hostname should be stored in environment variables. And depending on the environment, it can be e.g. `testing.adamant.im` for testing environment or `msg.adamant.im` for production. It will be fixed but not as part of this PR.

For now I'll create a constant in the test files to get rid of some repetitions.

src/components/__tests__/__snapshots__/AppNavigation.test.js.snap
bludnic: I think something changed in the Vuetify library. Looks like they added breaking spaces in the default slot of the `VIcon` component.

src/components/__tests__/AppNavigation.test.js
martiliones: I mean the test block is describing 'AppNavigation.vue':
```js
describe('AppNavigation.vue', () => {
```
and then you write "**it** badge should", which is not a right English sentence:
```js
it('badge ...'
```
It should be "it**s** badge", since "it" refers to 'AppNavigation.vue', but you can write "It (_AppNavigation.vue_) should hide the badge...":
```js
it('should hide the badge...'
```
The point is that the name of the method is part of the sentence. You can also just use `test` instead of `it`

martiliones (approved):

adamant-al (approved):

PR #450, opened 2023-03-31
replace options api for composition api

joebrock666:


src/views/Votes.vue
```diff
@@ -125,128 +125,155 @@ import AppToolbarCentered from '@/components/AppToolbarCentered'
...
-      if (!this.validateVotes()) {
+    }
+    const upVote = (id) => {
+      store.commit('delegates/upVote', id)
+    }
+    const downVote = (id) => {
+      store.commit('delegates/downVote', id)
+    }
+    const sendVotes = () => {
+      if (!validateVotes()) {
```
bludnic, line 201: The `validateVotes()` is called before declaration. Swap `sendVotes` and `validateVotes` function declarations.

src/views/Votes.vue
```diff
@@ -125,128 +125,155 @@ import AppToolbarCentered from '@/components/AppToolbarCentered'
 import Pagination from '@/components/Pagination'
 import { explorerUriToOnion } from '@/lib/uri'
 import DelegatesTable from '@/components/DelegatesTable/DelegatesTable'
+import { computed, onMounted, ref, reactive } from "vue"
+import { useStore } from "vuex"
+import { useI18n } from 'vue-i18n'
 
 export default {
```
bludnic, line 132: defineComponent

bludnic (changes requested):

src/views/Votes.vue
joebrock666: Alright, so I should use defineComponent function for both views and components?

src/views/Votes.vue
```diff
@@ -125,131 +125,153 @@ import AppToolbarCentered from '@/components/AppToolbarCentered'
...
+      totalVotes,
+      pages,
+      showPagination,
+      reviewButtonDisabled,
+      getExplorerUrl,
+      upVote,
+      downVote,
+      sendVotes,
+      validateVotes,
+      showConfirmationDialog
```
bludnic, line 271: Need to return `VOTE_REQUEST_LIMIT ` because it is used inside the `<template/>`

src/views/Votes.vue
bludnic: Yes.

`defineComponent` is just a type helper for TypeScript. It doesn't do anything.

bludnic (changes requested):

PR #461, opened 2023-05-01
chore(GitHub Actions): add `electron.yml` workflow

bludnic:
Add GitHub Workflows for building the Electron app for each platform:

- electron-mac.yml
- electron-linux.yml
- electron-windows.yml

https://trello.com/c/aQ85UVMM/332-add-github-workflow-for-electron-apps


build/osx/entitlements.mac.plist
```diff
@@ -6,5 +6,7 @@
     <true/>
     <key>com.apple.security.device.camera</key>
     <true/>
+    <key>com.apple.security.cs.allow-jit</key>
```
adamant-al, line 9: What's the reason to allow jit?
https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_cs_allow-jit

Earlier, builds worked without the this key.

notarize.js
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  if (process.env.SKIP_NOTARIZATION === 'true') {
+    console.log('SKIP_NOTARIZATION=true | Skipping the notarization"')
+    return
+  }
+
+  const appName = context.packager.appInfo.productFilename
+
+  return await notarize({
```
martiliones, line 18: remove `async` from function definition and return raw promise

martiliones (changes requested):

build/osx/entitlements.mac.plist
bludnic: This is required for notarization.
https://kilianvalkhof.com/2019/electron/notarizing-your-electron-application/

> One of the things Apple snuck in when adding the **notarization is the requirement of a hardened runtime** for your app, basically giving your application less rights by default.

> If you run a hardened runtime, it’s important to set at least two entitlements, “allow-unsigned-executable-memory” and **“allow-jit”**.

notarize.js
bludnic: Removed "await". I'll keep "async" to make it clear that the function must return a promise.

.github/workflows/electron-linux.yml
```diff
@@ -0,0 +1,37 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository 🛎️
+        uses: actions/checkout@v3
+
+      - name: Setup Node version 🍀
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16'
```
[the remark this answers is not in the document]
bludnic: I'll leave v16 for now

notarize.js
martiliones: why does it need to return a promise?

build/osx/entitlements.mac.plist
bludnic: Removed "allow-jit". The notarization works as well.

joebrock666 (approved):

adamant-al (approved):

PR #467, opened 2023-05-13
Bug/stay logged checkbox

joebrock666:
https://trello.com/c/sNhJMgNs/341-bug-stay-logged-in-checkbox-doesnt-work

### DIFF
--- a/src/views/Options.vue
+++ b/src/views/Options.vue
@@ -90,6 +90,7 @@
                 :label="$t('options.stay_logged_in')"
                 color="grey darken-1"
                 :input-value="stayLoggedIn"
+                v-model="stayLoggedIn"
                 readonly
                 density="comfortable"
                 hide-details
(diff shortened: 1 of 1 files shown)


src/views/Options.vue
```diff
@@ -90,6 +90,7 @@
                 :label="$t('options.stay_logged_in')"
                 color="grey darken-1"
                 :input-value="stayLoggedIn"
```
bludnic, line 92: `input-value` prop has been renamed to `v-model` in Vuetify 3, so you can remove this line.

Vuetify 2
https://v2.vuetifyjs.com/en/api/v-checkbox/#props-input-value
Vuetify 3
https://vuetifyjs.com/en/api/v-checkbox/#props-model-value

bludnic (changes requested):

bludnic (approved):

PR #478, opened 2023-06-01
Fix no spinner bug

joebrock666:
https://trello.com/c/0QIWLMyP/353-bug-no-spinner-when-loading-tx-list

### DIFF
--- a/src/views/Transactions.vue
+++ b/src/views/Transactions.vue
@@ -17,9 +17,9 @@
       >
         <v-list-item
           v-if="isRecentLoading"
-          style="position: absolute; top: 20px;"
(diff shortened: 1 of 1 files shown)


src/views/Transactions.vue
```diff
@@ -17,9 +17,9 @@
       >
         <v-list-item
           v-if="isRecentLoading"
-          style="position: absolute; top: 20px;"
+          style="position: relative; top: 20px;"
```
bludnic, line 20: Should remain absolute. You need to find a parent tag to which you will apply `position: relative`

bludnic (changes requested):
Check Trello comments

PR #493, opened 2023-06-16
fix(ChatForm): don't overflow the chat when typing a multiline message

bludnic:
https://trello.com/c/7jZEG9rt/228-bug-multiple-strings-hide-last-messages


adamant-al (changes requested):
UI bug, error in console.
See the Trello card.

adamant-al (approved):

PR #494, opened 2023-06-16
add stakecube exchange

joebrock666:


src/components/BuyTokensDialog.vue
```diff
@@ -47,6 +47,17 @@
 
             <v-list-item-title>Azbit</v-list-item-title>
           </v-list-item>
+
+          <v-list-item
+            avatar
+            @click="openLink('https://stakecube.net/app/exchange/adm_usdt?layout=pro&team=adm')"
+          >
+            <template #prepend>
+              <icon><stakeCube-icon /></icon>
```
bludnic, line 56: snake-case

joebrock666: You mean kebab-case?

bludnic: yes

bludnic (approved):

PR #498, opened 2023-06-28
change color to regular

joebrock666:


src/components/NodesTable/NodesTableItem.vue
```diff
@@ -168,11 +168,8 @@ export default {
 
 .v-theme--light {
   .nodes-table-item {
-    &__version {
```
bludnic, line 171: don't merge selectors, keep them separate

bludnic (changes requested):

src/components/NodesTable/NodesTableItem.vue
bludnic: The selectors may differ in the future, e.g. one selector may require a `margin` or `padding` and other not

joebrock666: fair enough

bludnic (approved):

PR #517, opened 2023-08-10
fix(`/transactions/ADM/:id`, updateStatus): app sending infinite requests

bludnic:
https://trello.com/c/tGvN9PhE/417-bug-after-adm-sending-the-app-sends-infinite-requests

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "adamant-im",
-  "version": "4.1.0",
+  "version": "4.1.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "adamant-im",
-      "version": "4.1.0",
+      "version": "4.1.1",
       "hasInstallScript": true,
       "license": "GPLv3",
       "dependencies": {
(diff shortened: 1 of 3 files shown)


adamant-al (changes requested):
1. Add to adamant-wallets/ADM:
```
 "txFetchInfo": {
      "newPendingInterval": 4000,
      "oldPendingInterval": 4000,
      "registeredInterval": 4000,
    },
```

2. Instead of 10000 use const:

`const DEFAULT_TX_FETCH_INTERVAL = 10000`

bludnic:
@adamant-al what about `newPendingAttempts` and `oldPendingAttempts`?

adamant-al:
> @adamant-al what about `newPendingAttempts` and `oldPendingAttempts`?

Do we actually use attempts for ADM?

It may be a good idea to add `const DEFAULT_TX_FETCH_ATTEMPTS = 5`.

bludnic:
> > @adamant-al what about `newPendingAttempts` and `oldPendingAttempts`?
> 
> Do we actually use attempts for ADM?
> 
> It may be a good idea to add `const DEFAULT_TX_FETCH_ATTEMPTS = 5`.

no, we don't

adamant-al (approved):

PR #537, opened 2023-10-06
feat: preserve message text in the input field

juliahermak:
[(after)Preserve message text in the input field.webm](https://github.com/Adamant-im/adamant-im/assets/122323015/d81bacf2-c49c-4ee9-8e2b-9ed027600b90)

### DIFF
--- a/src/components/AChat/AChatForm.vue
+++ b/src/components/AChat/AChatForm.vue
@@ -58,9 +58,16 @@ export default {
     clearInputValueOnSend: {
       type: Boolean,
       default: true
+    },
+    /**
+     * Message validator.
+     */
+    validator: {
+      type: Function,
+      required: true
     }
   },
-  emits: ['message', 'esc'],
+  emits: ['message', 'esc', 'error'],
   data: () => ({
     message: ''
   }),
@@ -122,10 +129,14 @@ export default {
   },
   methods: {
     submitMessage() {
-      this.$emit('message', this.message)
-      if (this.clearInputValueOnSend) {
+      const error = this.validator(this.message)
+      if (error === false) {
+        this.$emit('message', this.message)
         this.message = ''
+      } else {
+        this.$emit('error', error)
       }
+
       // Fix textarea height to 1 row after miltiline message send
       this.calculateInputHeight()
       this.focus()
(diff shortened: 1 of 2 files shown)


src/components/AChat/AChatForm.vue
```diff
@@ -119,7 +123,9 @@ export default {
   methods: {
     submitMessage() {
       this.$emit('message', this.message)
-      this.message = ''
+      if (this.clearInputValueOnSend === true) {
```
bludnic, line 126: shorter variant
```suggestion
      if (this.clearInputValueOnSend) {
```

src/components/Chat/Chat.vue
```diff
@@ -209,6 +209,7 @@
           :message-text="$route.query.messageText"
           @message="onMessage"
           @esc="replyMessageId = -1"
+          :clearInputValueOnSend="this.$store.state.balance < 0.001 === false"
```
bludnic, line 212: compare other side with `>=` and use shorter variant

```suggestion
          :clearInputValueOnSend="this.$store.state.balance >= 0.001"
```

bludnic: rename the prop name by using snake-case, just to keep the same style in the codebase

bludnic (changes requested):

bludnic:
[Feat: Preserve message text in the input field, if it was not sent](https://trello.com/c/HUvfBtph/363-feat-preserve-message-text-in-the-input-field-if-it-was-not-sent)

bludnic (approved):

adamant-al (changes requested):
I don't like the custom-case solution.
We have `validateMessage` method.
Also, consider moving magic numbers to constants.
Refactor code, if appropriate.

bludnic:
> I don't like the custom-case solution. We have `validateMessage` method. Also, consider moving magic numbers to constants. Refactor code, if appropriate.

I need to think about how to make the validation in one place and share it between Chat and AChatForm component, then I will came with a solution

bludnic:
- Added `validator` prop for `AChatForm`
- Replaced harcoded fee from constants

@adamant-al

src/components/Chat/Chat.vue
```diff
@@ -305,31 +307,33 @@ function getUserMeta(userId) {
 /**
  * Validate message before sending.
  * @param {string} message
- * @returns {boolean}
+ * @returns {string | false} If `false` then validation passed without errors.
  */
+const validationErrors = {
```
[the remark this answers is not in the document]
bludnic: moved above the docblock

PR #550, opened 2023-11-03
fix(ETH, ERC20): transfer funds doesn't work after `web3-eth` upgrade

bludnic:


src/store/modules/erc20/erc20-actions.js
```diff
@@ -14,26 +14,30 @@ const STATUS_INTERVAL = 25000
...
       .transfer(ethAddress, ethUtils.toWhole(amount, context.state.decimals))
       .encodeABI()
   }
 
-  return api.estimateGas(transaction).then((gasLimit) => {
-    gasLimit = increaseFee ? gasLimit * INCREASE_FEE_MULTIPLIER : gasLimit
-    transaction.gas = gasLimit
-    return transaction
-  })
+  const defaultGasLimit = await api.estimateGas(transaction)
```
adamant-al, line 35: 1. Rename `defaultGasLimit` ⟶ `gasLimit`
2. Remove `DEFAULT_ETH_TRANSFER_GAS` and `DEFAULT_ERC20_TRANSFER_GAS` from `constants/index.js`
3. `const gasLimit = api.estimateGas` or set to `CryptosInfo[${crypto}].defaultGasLimit` if unable to receive
4. Set `gasLimit` in transaction
5. Change `INCREASE_FEE_MULTIPLIER` from `2` to `1.5` with apropriate code changes

src/store/modules/erc20/erc20-actions.js
```diff
@@ -14,26 +14,30 @@ const STATUS_INTERVAL = 25000
...
   const contract = new EthContract(Erc20, context.state.contractAddress)
 
+  const nonce = await api.getTransactionCount(context.state.address)
+  const gasPrice = await api.getGasPrice()
+
   const transaction = {
     from: context.state.address,
     to: context.state.contractAddress,
     value: '0x0',
     // gasLimit: api.fromDecimal(DEFAULT_ERC20_TRANSFER_GAS), // Don't take default value, instead calculate with estimateGas(transactionObject)
```
adamant-al, line 27: Remove this and other commented code

src/store/modules/eth/actions.js
```diff
@@ -18,24 +18,25 @@ function storeEthAddress(context) {
...
+    // gasLimit: api.fromDecimal(DEFAULT_ETH_TRANSFER_GAS), // Don't take default value, instead calculate with estimateGas(transactionObject)
+    gasPrice,
+    nonce
   }
 
-  return api.estimateGas(transaction).then((gasLimit) => {
-    gasLimit = increaseFee
-      ? BigNumber(gasLimit).times(INCREASE_FEE_MULTIPLIER).toNumber()
-      : gasLimit
+  const defaultGasLimit = await api.estimateGas(transaction)
```
adamant-al, line 34: Same here. Can we move the code to base-actions and use both for eth and erc20?

src/store/modules/eth/actions.js
```diff
@@ -89,7 +90,7 @@ const createSpecificActions = (api) => ({
     void api.getGasPrice().then((price) => {
       // It is OK with London hardfork
```
adamant-al, line 91: Remove this comment

src/store/modules/eth/actions.js
```diff
@@ -89,7 +90,7 @@ const createSpecificActions = (api) => ({
...
       context.commit('gasPrice', {
-        gasPrice: price, // string type
+        gasPrice: Number(price), // string type
         fee: +(+utils.calculateFee(DEFAULT_ETH_TRANSFER_GAS, price)).toFixed(8) // number type, in ETH
```
adamant-al, line 94: Why we use `DEFAULT_ETH_TRANSFER_GAS` here?

src/store/modules/eth/actions.js
```diff
@@ -89,7 +90,7 @@ const createSpecificActions = (api) => ({
     void api.getGasPrice().then((price) => {
       // It is OK with London hardfork
       context.commit('gasPrice', {
-        gasPrice: price, // string type
+        gasPrice: Number(price), // string type
```
adamant-al, line 93: Remove the comment.

adamant-al (changes requested):

src/store/modules/erc20/erc20-actions.js
bludnic: 4. Not sure what you mean. The `gasLimit` is already set.

image

src/store/modules/eth/actions.js
bludnic: We cannot use `api.estimateGas()` here because we don't have the tx data at this moment. This action is triggered in the background and the purpose of this is to estimate the tx fee and show it in the UI.

image

src/store/modules/eth/actions.js
bludnic: I will create a task for refactoring `eth-base`. Now it's hard to understand what is happening here.
I think this function must be moved outside the store.

src/store/modules/erc20/erc20-actions.js
bludnic: 1, 3, 5 resolved

src/store/modules/eth/actions.js
adamant-al: Then create a tx in background and use `const gasLimit = api.estimateGas` or set to `CryptosInfo[${crypto}].defaultGasLimit` if unable to receive.
When we have no amount to send in the input field, calc as we send all the amount.
When we have no recipient in the input field, calc as we send to ourself.

src/store/modules/erc20/erc20-actions.js
adamant-al: 2- Remove, see my comment below.
4- It's done.

src/store/modules/eth/actions.js
bludnic: > When we have no amount to send in the input field, calc as we send all the amount.
When we have no recipient in the input field, calc as we send to ourself.

We can't do this without architectural change. Since we store input values (amount, recipient) in the component itself (SendFundsForm) we can't access this information from the store.

I suggest to skip this refactoring. Added the reference in https://trello.com/c/79uvtQU2/455-refactor-eth-erc20-store

adamant-al:
A successor: Refactor ETH/ERC20 store https://trello.com/c/79uvtQU2

adamant-al (approved):

adamant-al:
@martiliones @RealGoodProgrammer 
Review and merge.

bludnic:
[Update Web3 lib for Ethereum](https://trello.com/c/U9RxukKB/367-update-web3-lib-for-ethereum)

RealGoodProgrammer (approved):

PR #560, opened 2023-11-16
Decline transaction with same nonce (ETH, ERC20, LSK)

RealGoodProgrammer:
Display an error message when the user tries to send two transactions with equal `nonce`.

The last transaction `nonce` is saved in the `SessionStorage` right before sending the tokens.
Subsequent transactions are checked for the `nonce` equality. If the `nonces` are equal, an error message will be displayed.

In addition to the `nonce` transaction, the `expirationTime` is also saved. In rare situations the transaction may not reach the node (e.g. Network Error). We expect that `nonce` will change but it will not. The `expirationTime` must cover this edge case.

`expirationTime = coin.newPendingAttempts * coin.newPendingInterval`
(the values are taken from [adamant-wallets](https://github.com/Adamant-im/adamant-wallets/blob/master/assets/general/ethereum/info.json#L21))

- For ETH/ERC20: 80s
- For LSK: 45s

image


RealGoodProgrammer:
[Feat: Decline second crypto transfer if we get same Nonce](https://trello.com/c/HMtWwgV8/369-feat-decline-second-crypto-transfer-if-we-get-same-nonce)

adamant-al (changes requested):
This PR doesn't resolve the issue.

- You mark if the tx in process locally
- If you send BTC tx and restart the app, it will "think" that there is no pending txs

Consider to check current `nonce` for account.

Blockchains that have `nonce`: LSK, ETH
To research: BTC, DOGE, DASH

src/store/modules/eth-base/eth-base-actions.js
```diff
@@ -57,6 +57,9 @@ export default function createActions(config) {
     },
 
     sendTokens(context, { amount, admAddress, address, comments, increaseFee, replyToId }) {
+      if (context.state.isTransactionInProcess) return
```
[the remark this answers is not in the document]
RealGoodProgrammer: No, you're both wrong. I just don't want to throw this error twice. Check it for yourself

src/lib/txVerify.js
```diff
@@ -17,6 +17,59 @@ const TransactionInconsistentReason = {
   RECIPIENT_CRYPTO_ADDRESS_MISMATCH: 'recipient_crypto_address_mismatch'
 }
 
+export async function checkTxInProcess(api, context, withoutNonce = true, externalNonce = null) {
+  const fetchInfo = CryptosInfo[context.state.crypto].txFetchInfo
+  const TX_LOCK_TIME = fetchInfo.newPendingAttempts * fetchInfo.newPendingInterval
```
bludnic, line 22: According to the specification the `txFetchInfo` and its properties are optional. Take this into consideration and perform additional check.

image

src/lib/txVerify.js
```diff
@@ -17,6 +17,59 @@ const TransactionInconsistentReason = {
...
+
+  const currentNonce = context.rootGetters.getTransactionInProcess(context.state.crypto)
+
+  const expiration = Date.now()
```
bludnic, line 26: The name is confusing. I suggest to pick another name, e.g. `today`, `currentTimestamp`, or any other more meaningful name at your choice.

src/lib/errors.js
```diff
@@ -0,0 +1,12 @@
+export const ErrorCodes = {
+  TX_ALREADY_IN_PROCESS: 'TX_ALREADY_IN_PROCESS'
+}
+
+export class DuplicatedNonceError extends Error {
+  constructor(message, errorCode) {
```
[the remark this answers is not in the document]
bludnic: Also remove ErrorCodes above

src/lib/sessionStorage.js
```diff
@@ -0,0 +1,12 @@
+export const getFromSessionStorage = (key) => {
```
bludnic, line 1: Rewrite to TS. An example how to type it properly with generic:
https://github.com/Adamant-im/adamant-im/blob/master/src/lib/nodes/storage.ts#L17

src/lib/txVerify.js
```diff
@@ -17,6 +19,37 @@ const TransactionInconsistentReason = {
   RECIPIENT_CRYPTO_ADDRESS_MISMATCH: 'recipient_crypto_address_mismatch'
 }
 
+export function checkIsTxInProcess(coin, nonce = null) {
```
bludnic, line 22: Move function to separate file. Rewrite to TS.

bludnic (changes requested):

adamant-al (changes requested):
We've chosen another approach.
See Feat: Decline second crypto transfer if we get same Nonce https://trello.com/c/HMtWwgV8

bludnic:
Will be implemented in https://github.com/Adamant-im/adamant-im/pull/578

PR #566, opened 2023-11-27
Fix ESLint errors and Prettify entire codebase

bludnic:
Better to review commit by commit. It will be hard to see what changed because of the prettier formatting.

- [x] Fixed ESLint errors
- [x] Prettify entire codebase
- [x] Run `eslint` and `prettier` on lint-staged
- [x] Enabled absolute imports in SCSS files (`@/*`)
- [x] Fixed `BigInt is not defined` error in IDE
- [x] Added `commitlint` with conventional config
- [x] Added `npm run typecheck` script
- [x] Include `vite-*.config.ts` files in `tsconfig`


bludnic:
> Can we add this plugin? I remember you asking somebody to order imports in a pull request
https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md

Yes, it's possible. I will add it.

bludnic:
Actually `eslint-plugin-import` doesn't sort imports. I added it anyway so we can catch other issues related to imports.

For sorting, I checked `eslint-plugin-perfectionist`. It works but I don't like how it sorts the imports (tested all styles). Ideally I would like to see node_modules imports at the top, and the projects imports at the bottom. Looks like this plugin can't do this.

There a files that imports Vuex store directly. Changing the imports order may break the app. I will handle this in other task.

bludnic:
@martiliones Ok. Anyway I will leave sorting imports for another task because of reasons mentioned before.

PR #571, opened 2023-12-21
fix(nodes): can't login when no online nodes available

bludnic:


bludnic:
[Can't login in the app if all nodes are unavailable](https://trello.com/c/0TlZWvFh/474-cant-login-in-the-app-if-all-nodes-are-unavailable)

PR #596, opened 2024-02-02
fix: STORJ/GT order

RealGoodProgrammer:


adamant-al:
Close it.
Order update will come with next fetch from adamant-wallets.

bludnic:
See https://github.com/Adamant-im/adamant-wallets/pull/71

PR #604, opened 2024-03-12
fix: hide icon in send funds form

juliahermak:
Screenshot 2024-03-12 01 06 03


juliahermak:
[Bug: Sending coins from chat: the list of coins isn't opening](https://trello.com/c/1TbxOke5/502-bug-sending-coins-from-chat-the-list-of-coins-isnt-opening)

PR #605, opened 2024-03-12
Feat/cycle through commands

PaulDremanovich:
https://trello.com/c/ZnpoHark/364-feat-cycle-through-commands-with-arrow-keys

----

You can use ADAMANT for managing bots.

Every bot command starts with a slash /.

In case if:

The input field is focused and empty
Arrow keys Up and Down:

Cycle through the previous sent commands
Only commands in local storage, don’t upload messages from blockchain
Put a cursor to the end of text
Up is the previous command, Down is next
Down does nothing if it pressed before the first Up
Up does nothing if we reached the very first command in a chat
Cycling state is paused when:

A command (the input filed) is modified during cycling
Cycling is restarted when:

A user clears the input field

[Screencast from 14.03.2024 18:51:48.webm](https://github.com/Adamant-im/adamant-im/assets/160286261/b1d88ef8-e20d-47c4-adb5-e0eabc8dfbce)


src/components/icons/cryptos/ManaIcon.vue
RealGoodProgrammer, line 1: It's happened automatically after merge

src/components/AChat/AChatForm.vue
```diff
@@ -232,6 +246,30 @@ export default {
...
+            if (this.botCommandIndex > 0) {
+              this.botCommandIndex--
+            }
+          } else if (this.botCommandIndex < maxIndex) {
+            this.botCommandIndex++
+          }
+        }
+
+        this.message = commands[this.botCommandIndex] || ''
+      }
```
bludnic, line 272: Avoid using deep nested if/else conditions. Instead try using early return. This will improve the code readability.

E.g.:

```js
if (!this.botCommandSelectionMode) {
   return;
}

const commands = this.$store.getters['botCommands/getCommandsHistory'](this.partnerId)
// ...
```

src/components/AChat/AChatForm.vue
```diff
@@ -154,6 +156,12 @@ export default {
     onKeyCommand: function (event) {
       if (event.ctrlKey && event.shiftKey && event.code === 'Digit1') {
         this.openElement()
+      } else if (event.code === 'ArrowUp' || event.code === 'ArrowDown') {
+        this.selectCommand(event.code)
+        event.preventDefault()
```
bludnic, line 161: Better to move `event.preventDefault()` at the top of the function body.

bludnic (changes requested):

src/components/AChat/AChatForm.vue
PaulDremanovich: if i'll move preventDefault() at the top of function body it will totally break input's behavior.
So i'm gonna leave it unchanged.

src/components/AChat/AChatForm.vue
```diff
@@ -232,6 +246,31 @@ export default {
...
+        }
+      } else {
+        if (direction === 'ArrowUp') {
+          if (this.botCommandIndex > 0) {
+            this.botCommandIndex--
+          }
+        } else if (this.botCommandIndex < maxIndex) {
+          this.botCommandIndex++
+        }
+      }
```
bludnic, line 271: Can you simplify it as well by using early return?

bludnic (changes requested):

adamant-al:
@PaulDremanovich 
- Fill the description for each PR.
- Add a name of a Trello card and a link to it
- Add screenshot if it's informative

bludnic (approved):

src/store/modules/bot-commands/bot-commands-mutations.ts
```diff
@@ -0,0 +1,13 @@
+import { MutationTree } from 'vuex'
+import { BotCommand, BotCommandsState } from '@/store/modules/bot-commands/types.ts'
+
+export const mutations: MutationTree<BotCommandsState> = {
+  addCommand(state, botCommand: BotCommand): void {
+    const commandValue = botCommand.command.trim()
+    if (!state.commands[botCommand.partnerId]) {
```
adamant-al, line 7: Can we set `botCommands = state.commands[botCommand.partnerId]` to make code simplier?

adamant-al:
```
addCommand(state, botCommand: BotCommand): void {
  const commandValue = botCommand.command.trim();
  let botCommands = state.commands[botCommand.partnerId];

  if (!botCommands) {
    state.commands[botCommand.partnerId] = [commandValue];
  } else if (!botCommands.includes(commandValue)) {
    botCommands.push(commandValue);
  }
}
```

adamant-al: May be we also can init `state.commands[botCommand.partnerId] = []` somewhere on start and keep only

```
addCommand(state, botCommand: BotCommand): void {
  const commandValue = botCommand.command.trim();
  let botCommands = state.commands[botCommand.partnerId];

  if (!botCommands.includes(commandValue)) {
    botCommands.push(commandValue);
  }
}
```

PaulDremanovich: > May be we also can init `state.commands[botCommand.partnerId] = []` somewhere on start and keep only
> 
> ```
> addCommand(state, botCommand: BotCommand): void {
>   const commandValue = botCommand.command.trim();
>   let botCommands = state.commands[botCommand.partnerId];
> 
>   if (!botCommands.includes(commandValue)) {
>     botCommands.push(commandValue);
>   }
> }
> ```

It seems to me that this will be dangerous, because... There is a life cycle problem, namely at what point to call initialization? We can call initialization at start. But what about the new chat? What about updating the page? Whether the method will be called or not. One way or another, you will need to check that the array of commands already exists.
It seems to me inappropriate to increase the complexity of the life cycle in order to save one line of code.

adamant-al (approved):

bludnic (approved):

PR #609, opened 2024-03-16
fix: emojis are not visible on iOS Chrome

RealGoodProgrammer:
Fix for hidden emojis selector on Chrome (iOS
image (8)

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "private": true,
   "scripts": {
     "serve": "vite preview --config vite-pwa.config.ts",
-    "dev": "vite --config vite-pwa.config.ts",
+    "dev": "vite --host 0.0.0.0 --config vite-pwa.config.ts",
     "serve:test": "cross-env ADM_CONFIG_FILE=test vue-cli-service serve",
     "wallets:generate": "npm run wallets:data:generate && npm run wallets:types:generate",
     "wallets:data:generate": "node scripts/wallets.mjs",
--- a/package.json
+++ b/package.json
@@ -7,7 +7,8 @@
   "private": true,
   "scripts": {
     "serve": "vite preview --config vite-pwa.config.ts",
-    "dev": "vite --host 0.0.0.0 --config vite-pwa.config.ts",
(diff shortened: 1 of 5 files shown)


bludnic:
@RealGoodProgrammer Attach the Trello card to the PR

package.json
```diff
@@ -7,7 +7,7 @@
   "private": true,
   "scripts": {
     "serve": "vite preview --config vite-pwa.config.ts",
-    "dev": "vite --config vite-pwa.config.ts",
+    "dev": "vite --host 0.0.0.0 --config vite-pwa.config.ts",
```
bludnic, line 10: Why is this needed?

src/components/EmojiPicker.vue
```diff
@@ -68,6 +68,8 @@ export default defineComponent({
 
 .emoji-picker {
   border-radius: 8px;
+  position: absolute;
+  bottom: 0;
```
bludnic, line 72: Does `<v-menu location="top"` works without these styles?

bludnic (changes requested):

bludnic (approved):

src/components/EmojiPicker.vue
RealGoodProgrammer: Not in Google Chrome on iOS

bludnic (approved):

bludnic:
[UI: Add emoji and move attach to the input field (copied from iOS)](https://trello.com/c/x5HC9Uyj/426-ui-add-emoji-and-move-attach-to-the-input-field-copied-from-ios)

PaulDremanovich:
Why do you have the positioning property implemented if you keep hardcoding it early? Perhaps you just need to change the style?

RealGoodProgrammer:
> Why do you have the positioning property implemented if you keep hardcoding it early? Perhaps you just need to change the style?

For the case when it should be static.  Look at the condition. The class will be applied only when the 'absolute' value is passed to it in the property

PaulDremanovich (approved):

PR #640, opened 2024-06-30
`Object.groupBy` is not a function in older browsers

bludnic:


bludnic:
[Object.groupBy is not a function in Tor browser](https://trello.com/c/R5hjSyzM/562-objectgroupby-is-not-a-function-in-tor-browser)

PR #646, opened 2024-07-03
Fix WASM error

bludnic:


bludnic:
[WebAssemply error in surge build](https://trello.com/c/lUygdHCy/518-webassemply-error-in-surge-build)

PR #674, opened 2024-10-22
Feat/support kly qrcode

NikIvv:
Read QR from an image file created in Klayr wallet.

### DIFF
--- a/src/components/SendFundsForm.vue
+++ b/src/components/SendFundsForm.vue
@@ -125,7 +125,7 @@
         v-if="isTextDataAllowed"
         v-model="textData"
         class="a-input"
-        :label="klayrOptionalMessage ? klayrOptionalMessage : textDataLabel"
+        :label="textDataLabel"
         variant="underlined"
         counter
         maxlength="64"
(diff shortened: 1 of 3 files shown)


src/components/SendFundsForm.vue
```diff
@@ -701,13 +703,16 @@ export default {
      */
     onScanQrcode(uri) {
       const recipient = parseURIasAIP(uri)
```
bludnic, line 705: I suggest to create a separate function for parsing KLY based URIs. For example:

```ts
function parseKlyURI(uri: string) {
  // ...
  return { address, amount,  }
}
```

Then you can create a function that will unify the result of both parsers:

```ts
function parseURI() {
  if (isKlyUri) {
    return parseKlyURI()
  }

  return parseURIasAIP() // ADM based URI
}
```

bludnic (changes requested):

bludnic:

bludnic:
[Feat: Enable KLY QR codes from their official app (copied from iOS)](https://trello.com/c/PPllsFpR/603-feat-enable-kly-qr-codes-from-their-official-app-copied-from-ios)

src/lib/uri.js
```diff
@@ -18,11 +20,69 @@ export function getAddressBarURI() {
...
+  return { address, crypto: Cryptos.KLY, params, protocol: KLAYR_WALLET }
+}
+
 /**
  * Parse info from an URI containing a cryptocurrency address
  * Complies with AIP-2, AIP-8, AIP-9
  * Sample: https://msg.adamant.im?address=U9821606738809290000&label=John+Doe&amount=1.12&message=Buy+a+beer
- * @param {string} uri URI. Default is address bar or argv[].
+ * @param {string} URI's origin
+ * @param {string} URI's query parameters
```
bludnic, line 85: There is only one param `parseURIasAIP(uri)`

src/lib/uri.js
```diff
@@ -18,11 +20,69 @@ export function getAddressBarURI() {
...
+    return key && value
+      ? {
+          ...accum,
+          [key]: window.decodeURIComponent(
+            value.includes('+') ? value.replace(/\+/g, ' ') : value
+          )
+        }
+      : accum
+  }, Object.create(null))
+}
```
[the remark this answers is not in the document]
NikIvv: This change, in turn, make us to do further changes (not big ones) in parseURIasAIP(). If we need to get rid of formQueryParamsObject function completely, I'll do that. Please confirm.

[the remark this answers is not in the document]
bludnic: You can transform `paramsFromUrl` into an object so it should behave as the previous implementation. Or, is there something else I miss?

src/lib/uri.js
```diff
@@ -74,15 +74,14 @@ function parseKlyURI(query) {
     address = params.recipient || ''
   }
 
-  return { address, crypto: Cryptos.KLY, params, protocol: KLAYR_WALLET }
+  return { address, crypto: Cryptos.KLY, params, protocol: Cryptos.KLY }
```
bludnic, line 77: Cryptos.KLY -> Cryptos.KLY.toLowerCase()

bludnic (approved):

PR #680, opened 2024-11-10
feat/checkbox-check-uncheck-all-nodes

NikIvv:


src/components/nodes/adm/AdmNodesTable.vue
```diff
@@ -29,23 +29,33 @@ export default defineComponent({
...
     const admNodes = computed(() => {
       const arr = store.getters['nodes/adm']
 
       return [...arr].sort(sortNodesFn)
     })
 
+    const isAllChecked = ref(false)
+
+    const isAllNodesEnabled = !admNodes.value.some(node => node.active === false)
+    if (isAllNodesEnabled) isAllChecked.value = true
```
bludnic, line 42: You don't need to define a new state (line 39), since the value can be computed from the actual Vuex state

```suggestion
    const isAllChecked = computed(() => admNodes.value.every(node => node.active))
```

src/components/nodes/adm/AdmNodesTable.vue
```diff
@@ -29,23 +29,33 @@ export default defineComponent({
...
+
+    watch(isAllChecked, (value) => {
+      admNodes.value.forEach((admNode) => {
+        if (admNode && admNode.active !== value) {
+          store.dispatch('nodes/toggle', {...admNode, active: value})
+        }
+      })
+    })
```
bludnic, line 50: 1. `watch` no more needed here (check previous comment). Use a function instead.

2. Add a new action `nodes/toggleAll` in Vuex module:

```js
const toggleAll = (active) => {
   store.dispatch('nodes/toggleAll', { active })
}
```

3. And now in the `<template>` instead of `v-model`:

```vue
<NodesTableHead modelValue="isAllChecked" @update:modelValue="toggleAll" />
```

Note: If you want to keep the `v-model` approach, define the `computed` with a getter and setter. Example:

https://github.com/Adamant-im/adamant-im/blob/208cb916ff263f428db6dd766fd7d292dd7ad33d/src/components/CurrencySwitcher.vue#L43-L53

src/components/nodes/adm/AdmNodesTable.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <NodesTableContainer>
-    <NodesTableHead />
+    <NodesTableHead v-model="isAllChecked" />
```
bludnic, line 3: I would consider to use a slot instead, but your approach is also fine:

```vue
<NodesTableHead>
  <NodeStatusCheckbox v-model="isAllEnabled" />
</NodesTableHead>
```

bludnic (changes requested):

src/components/nodes/adm/AdmNodesTable.vue
bludnic: Resolve if you leave as is

src/components/nodes/adm/AdmNodesTable.vue
NikIvv: I defined not a new state, but a variable used in v-model, see NodesTableHead. That additional check is made for the case of the new tab switch with all nodes have been already checked before.

NikIvv:
[Checkbox to check/uncheck all Nodes](https://trello.com/c/DdP9Z37n/624-checkbox-to-check-uncheck-all-nodes)

src/components/nodes/services/ServiceNodesTable.vue
```diff
@@ -44,10 +49,30 @@ export default defineComponent({
       return [...arr].sort(sortCoinNodesFn)
     })
 
+    const isAllNodesChecked = computed({
+      get() {
+        return !nodes.value.some(node => node.active === false)
```
bludnic, line 54:
```suggestion
        return nodes.value.every(node => node.active)
```

simplify

src/components/nodes/services/ServiceNodesTable.vue
```diff
@@ -44,10 +49,30 @@ export default defineComponent({
...
+      },
+      set(value) {
+        nodes.value.forEach((node) => {
+          if (node && node.active !== value) {
```
bludnic, line 58: No need to check truthy on `node` since it is always defined.

src/components/nodes/services/ServiceNodesTable.vue
```diff
@@ -44,10 +49,30 @@ export default defineComponent({
...
+            const { label, url } = node
+            store.dispatch('services/toggle', {type: label, url, active: value})
```
bludnic, line 60: Create an action in Vuex: `store.dispatch('services/toggleAll', active)`

src/components/nodes/components/NodesTableHead.vue
```diff
@@ -37,22 +55,27 @@ export default {
...
     label: {
       type: String
-    }
+    },
+    indeterminate: {
+      type: Boolean
+    },
   },
-  setup() {
+  setup(props, { emit }) {
```
bludnic, line 63: Define `emits: ['update:modelValue'],` above. You must see a warning in the console without it.

bludnic (changes requested):

src/components/nodes/components/NodesTableHead.vue
bludnic: I don't see the change. The `emits` property is not defined.

See example: https://github.com/Adamant-im/adamant-im/blob/c6c73452bbb1b8fbb0f907c611d5672442ad871a/src/components/ShareURIDialog.vue#L57

bludnic (changes requested):

src/components/nodes/adm/AdmNodesTableItem.vue
```diff
@@ -119,4 +120,10 @@ export default {
 .amd-nodes-table-item {
   line-height: 14px;
 }
+
+.amd-nodes-table-item {
+  &__column--status {
+    max-width: 84px;
+  }
```
bludnic, line 127: Don't duplicate `.amd-nodes-table-item` class

bludnic: Approved in advance

bludnic (changes requested):

bludnic (approved):

PR #681, opened 2024-11-15
Update doge indexer healthcheck endpoint

NikIvv:


NikIvv:
[Update Doge Indexer healthcheck endpoint](https://trello.com/c/cz8O7Z3g/625-update-doge-indexer-healthcheck-endpoint)

PR #685, opened 2024-11-24
fix: usdt tx fee

NikIvv:


NikIvv:
[Bug: USDT tx fee in USDT and it's wrong](https://trello.com/c/R9ep2beI/617-bug-usdt-tx-fee-in-usdt-and-its-wrong)

bludnic (approved):

PR #690, opened 2024-12-02
fix(NodesOfflineDialog.vue): fix paddings

bludnic:


bludnic:

PR #696, opened 2024-12-26
fix: bug kly amount more than n characters

NikIvv:


src/components/SendFundsForm.vue
```diff
@@ -527,7 +527,14 @@ export default {
           () =>
             isErc20(this.currency)
               ? this.ethBalance >= this.transferFee || this.$t('transfer.error_not_enough_eth_fee')
-              : true
+              : true,
+          (v) => {
+            const isKlyTransfer = this.currency === Cryptos.KLY
+            if (!isKlyTransfer) return true
+            const MAX_UINT64 = BigInt('18446744073709551615')
```
bludnic, line 534: Can you import it from Klayr lib?
https://github.com/klayrHQ/klayr-sdk/blob/development/elements/validator/src/constants.ts#L19

src/components/SendFundsForm.vue
```diff
@@ -527,7 +527,14 @@ export default {
...
+            const isKlyTransferAllowed = isKlyTransfer && this.transferFee && v < MAX_UINT64
```
bludnic, line 535: 1. Redundant check on `isKlyTransfer `. You already used early return on line 533.

2. What's the purpose on checking `this.transferFee`?

3. I think `MAX_UINT64` is represented in Beddows, while the amount (`v`) is represented in KLY. If so, you must first convert `v` into Beddows and then compare it with `MAX_UINT64`. Check `convertklyToBeddows` helper.

bludnic (changes requested):

src/components/SendFundsForm.vue
bludnic: 1 KLY = 100,000,000 Beddows

src/components/SendFundsForm.vue
NikIvv: No, otherwise I would import it.

src/components/SendFundsForm.vue
NikIvv: 2. "Handle the runtime error when estimating [fee]. Return 0 fee if error catched" In normal cases this.transferFee cannnot be zero.

src/components/SendFundsForm.vue
bludnic: Try this

```js
import { MAX_UINT64 } from '@klayr/validator'
```

src/lib/klayr/klayr-utils.ts
```diff
@@ -232,32 +232,37 @@ type EstimateFeeParams = {
...
 export function estimateFee(params?: EstimateFeeParams) {
-  const {
-    amount = '1',
-    keyPair = KLY_DEMO_ACCOUNT.keyPair,
-    recipientAddress = KLY_DEMO_ACCOUNT.address,
-    data = '',
-    nonce = 0,
-    isNewAccount
-  } = params || {}
+  try {
```
bludnic, line 235: 1. Use try/catch inside the `kly/fee` getter:

https://github.com/Adamant-im/adamant-im/blob/fe08d661dcb21eb6a253791f0246fafb72adf77b/src/store/modules/kly/kly-getters.js#L8

2. Rollback changes in the current file

bludnic (changes requested):

src/store/modules/kly/kly-getters.js
```diff
@@ -5,12 +5,17 @@ export default {
...
+      return estimateFee({
+        amount,
+        data,
+        isNewAccount,
+        nonce: state.nonce
+      })
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    } catch (error) {
+      return 0
+    }
```
bludnic, line 18: Just omit the `error` argument from catch block. ESLint comment not needed.

```suggestion
    } catch {
      return 0
    }
```

bludnic (changes requested):

bludnic (approved):

PR #697, opened 2025-01-11
fix(KVS): wrong wallet address stored in KVS while logging into another account

bludnic:
This scenario may occur in-between the `GET /states/get` and `POST /state/store` requests, if the first request is pending while the user logs into another account. A check has been added to interrupt the second request.
https://trello.com/c/0kgxXGmj/639-wrong-wallet-address-stored-in-kvs-while-logging-into-another-account


bludnic:
[Bug: Wrong wallet address stored in KVS while logging into another account](https://trello.com/c/0kgxXGmj/639-bug-wrong-wallet-address-stored-in-kvs-while-logging-into-another-account)

PR #700, opened 2025-01-26
fix: infinite queue of requests while no connection

PaulDremanovich:


PaulDremanovich:
[Bug: Infinite queue of requests while no connection](https://trello.com/c/2zfCSUSC/623-bug-infinite-queue-of-requests-while-no-connection)

src/views/Wallets.vue
```diff
@@ -153,16 +153,19 @@ export default defineComponent({
...
+    let timer = ref<Timer>()
+    ;(async () => {
+      const updateBalances = async () => {
+        await store.dispatch('updateBalance', {
+          requestedByUser: true
+        })
+        timer.value = setTimeout(() => updateBalances(), BALANCE_UPDATE_INTERVAL_MS)
+      }
+      await updateBalances()
+    })()
```
bludnic, line 165: `updateBalances` is invoked right inside `setup()` method. Better to put it inside a hook, e.g. `onMount()`

bludnic (changes requested):

src/views/Wallets.vue
S-FrontendDev: i agree, it will be more readable

bludnic: Check as well maybe you can find a hook on https://vueuse.org/ what simplify that logic. We already have `@vueuse/core` as a dependency.

bludnic (approved):

S-FrontendDev (approved):

PR #701, opened 2025-01-27
fix: hide signal messages in chats

S-FrontendDev:


bludnic:
I think the issue is that we push the message locally to the chat, so if you refresh the page, the Signal Message will disappear. This means that the nodes don't return Signal Messages neither via WS or REST.

@yoxira can you confirm that these endpoints don't return Signal Messages?

- WS `socket('newMessage', (transaction) => {})`
- GET `/api/chats/get`

src/store/plugins/socketsPlugin.js
```diff
@@ -14,7 +15,10 @@ function subscribe(store) {
       // All transactions we get via socket are shown in chats, including ADM direct transfers
       // Currently, we don't update confirmations for direct transfers, see getChats() in adamant-api.js
       // So we'll update confirmations in getTransactionStatus()
-      store.dispatch('chat/pushMessages', [decoded])
+
+      if (transaction.asset?.chat?.type !== MessageType.SIGNAL_MESSAGE) {
+        store.dispatch('chat/pushMessages', [decoded])
+      }
```
bludnic, line 21: See https://github.com/Adamant-im/adamant-im/pull/701#issuecomment-2617533368

bludnic (changes requested):

src/store/plugins/socketsPlugin.js
S-FrontendDev: We are getting this messages from socket

yoxira:
@bludnic 

- `/api/chats/get` doesn't return signal messages unless you directly use `?type=3` parameter.
-  Node doesn't emit `newMessage` event. It's a client event (See https://github.com/Adamant-im/adamant-im/blob/dev/src/lib/sockets.js#L152). `newTrans` event **does** return all transactions by default, including signal messages.

bludnic (approved):

PR #704, opened 2025-01-30
fix: line separator to line feed

S-FrontendDev:


S-FrontendDev:
[Treat Line Separator as a Line Feed](https://trello.com/c/bTMNw7j6/631-treat-line-separator-as-a-line-feed)

bludnic (approved):

PR #707, opened 2025-02-04
Feat/svg mdi

S-FrontendDev:


S-FrontendDev:
[Import MDI icons as SVG instead of using CSS font](https://trello.com/c/5URNzGgy/615-import-mdi-icons-as-svg-instead-of-using-css-font)

NikIvv (approved):

src/components/AChat/AChatMessageActionsDropdown.vue
```diff
@@ -60,7 +62,7 @@ export default defineComponent({
       emit('open:change', state, props.transaction)
     }
 
-    return { t, classes, toggleMenu, emit }
+    return { t, classes, mdiChevronDown,toggleMenu }
```
bludnic, line 65: prettify, missing space after comma

src/components/AppSnackbar.vue
```diff
@@ -19,14 +19,21 @@
...
 
 <script>
+import { mdiClose } from '@mdi/js'
+
 export default {
+  setup() {
+    return {
+      mdiClose
+    }
+  },
```
bludnic, line 36: I'm not sure if it's a good idea to mix Composition API with Options API. Can you show an example from Vue docs that this approach is fine? @S-FrontendDev

S-FrontendDev: Project already had a lot of components with mix

And i think its good way to init image variable

And this is a good foundation for future refactoring, since in the future you will have to transfer less code under the setup block

You can check article by vue - https://vuejs.org/api/composition-api-setup.html

They provide us mix variant. Because they use mounted hook from options api

bludnic (approved):

adamantmm:
@bludnic @NikIvv 
Please add short descriptions for PRs.
It's not clear what this PR changes and why.

PR #711, opened 2025-02-11
fix: line separator next line space

S-FrontendDev:


S-FrontendDev:
[Treat Line Separator as a Line Feed](https://trello.com/c/bTMNw7j6/631-treat-line-separator-as-a-line-feed)

src/lib/markdown.js
```diff
@@ -72,12 +72,15 @@ export function removeFormats(text = '') {
 
 export function formatMessage(text = '') {
   const node = document.createElement('div')
-  const textWithSymbol = text.replace(/\n/g, '↵ ')
+
+  const replaceValue = '↵ '
```
bludnic, line 76: 1. Choose more meaningful name, like `LINE_BREAK_SYMBOL`.
2. Move outside function, near `const LINE_SEPARATOR`
3. Add a comment, describing the purpose of that constant

bludnic (changes requested):

src/lib/markdown.js
adamantmm: May be `LINE_BREAK_VISUAL`

bludnic (approved):

adamant-al:
Good work

PR #713, opened 2025-02-12
fix: return back icons

S-FrontendDev:


S-FrontendDev:
[Import MDI icons as SVG instead of using CSS font](https://trello.com/c/5URNzGgy/615-import-mdi-icons-as-svg-instead-of-using-css-font)

bludnic (approved):

PR #714, opened 2025-02-12
fix: adm node status

S-FrontendDev:


S-FrontendDev:
[Bug: When ADM proxy node is not available, it's shown as Not supported/ Outdated API version](https://trello.com/c/YrJmJdyg/658-bug-when-adm-proxy-node-is-not-available-its-shown-as-not-supported-outdated-api-version)

bludnic (approved):

PR #722, opened 2025-02-19
fix: display fee in ETH

S-FrontendDev:


S-FrontendDev:
[Bug: USDT tx fee in USDT and it's wrong](https://trello.com/c/R9ep2beI/617-bug-usdt-tx-fee-in-usdt-and-its-wrong)

bludnic (approved):

PR #723, opened 2025-02-19
fix: send a healthcheck request only if such request is not currently being executed

PaulDremanovich:
[Screencast from 20.02.2025 00:19:59.webm](https://github.com/user-attachments/assets/a4528085-3eaa-4bf9-af35-90f09e25866e)


PaulDremanovich:
[Bug: Infinite queue of requests while no connection](https://trello.com/c/2zfCSUSC/623-bug-infinite-queue-of-requests-while-no-connection)

bludnic (approved):

PR #725, opened 2025-02-20
feat: estimate currency commission in tx details

S-FrontendDev:


S-FrontendDev:
[Feat: In the Tx details screen show Fee estimate in USD](https://trello.com/c/IWgP4aVD/689-feat-in-the-tx-details-screen-show-fee-estimate-in-usd)

PaulDremanovich (approved):

src/components/transactions/TransactionTemplate.vue
```diff
@@ -289,6 +289,33 @@ export default defineComponent({
...
+
+      return timestampInSec(props.crypto, transaction.value.timestamp!)
+    })
+
+    const calculatedFee = computed(() => {
+      const commissionTokenLabel = props.feeCrypto ?? props.crypto;
+      const tokenFee = typeof props.fee === 'number' ? formatAmount(props.fee) + ` ${commissionTokenLabel}` : placeholder.value;
+
+      if (props.fee && calculatedTimestampInSec.value) {
+        const usdTokenRate = store.state.rate.historyRates?.[calculatedTimestampInSec.value]?.[`${commissionTokenLabel}/USD`];
```
bludnic, line 305: Check, maybe there is a getter for this

src/components/transactions/TransactionTemplate.vue
```diff
@@ -289,6 +289,33 @@ export default defineComponent({
...
 
+    const calculatedTimestampInSec = computed(() => {
+      if (!transaction.value) {
+        return null;
+      }
+
+      return timestampInSec(props.crypto, transaction.value.timestamp!)
+    })
+
+    const calculatedFee = computed(() => {
```
bludnic, line 300: Can you simplify? Seems a lot of logic inside that computed.

S-FrontendDev: i can add some computed var's, but for what? We don't use info from this var outside. But i will try to simplify it

S-FrontendDev: i added store getter for currency amount, a little bit simpler. And we can reuse this getter later

bludnic (approved):

PR #726, opened 2025-02-21
fix: tx list broken

S-FrontendDev:


S-FrontendDev:
[Bug: DASH and BTC tx list is broken](https://trello.com/c/Lbe8bQen/693-bug-dash-and-btc-tx-list-is-broken)

bludnic (approved):

PR #727, opened 2025-02-21
fix: change shouldUpdate callback for wallet status

S-FrontendDev:
wallets/getVisibility getter return undefined, need to recheck visibility after all wallets init

### DIFF
--- a/src/store/utils/coinUpdatesGuard.ts
+++ b/src/store/utils/coinUpdatesGuard.ts
@@ -5,11 +5,5 @@ export default function shouldUpdate(isVisibleCallback: Callback) {
     return true
   }
 
-  const isVisible = isVisibleCallback()
-
-  if (isVisible === undefined) {
-    return true;
-  }
-
-  return isVisible
+  return isVisibleCallback()
 }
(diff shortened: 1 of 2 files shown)


S-FrontendDev:
[Bug: ETH transfer fee is not specified](https://trello.com/c/c3x2SJcB/690-bug-eth-transfer-fee-is-not-specified)

src/store/utils/coinUpdatesGuard.ts
```diff
@@ -5,5 +5,11 @@ export default function shouldUpdate(isVisibleCallback: Callback) {
...
   }
 
-  return isVisibleCallback()
+  const isVisible = isVisibleCallback()
+
+  if (isVisible === undefined) {
+    return true;
+  }
+
+  return isVisible
```
bludnic, line 14: This fix doesn't seems right for me. You should check why the getter returns `undefined` when it should return `true`.

https://github.com/Adamant-im/adamant-im/blob/d525f92bcb3ba6fb1bcf2c9269812973504259a0/src/store/modules/eth/actions.js#L78

bludnic: If, e.g. ETH is enabled on Wallets list page the getter should return `true`

image

S-FrontendDev: Because it doesn't have time to initialize

Okay, I'll look into the logic for initializing wallets

bludnic (approved):

PR #729, opened 2025-02-24
feat: add HOURGLASS icon for not estimated rate

S-FrontendDev:


S-FrontendDev:
[Bug: ETH transfer fee is not specified](https://trello.com/c/c3x2SJcB/690-bug-eth-transfer-fee-is-not-specified)

bludnic (approved):

PR #730, opened 2025-02-25
chore: update marked and dompurify versions and verify functionality

skranee:


package.json
```diff
@@ -75,7 +75,7 @@
     "js-base64": "^3.7.7",
     "js-md5": "^0.8.3",
     "lodash": "^4.17.21",
-    "marked": "^14.1.0",
+    "marked": "^15.0.7",
```
bludnic, line 78: Although the library supports global scope, I suggest using the instance approach: https://marked.js.org/using_advanced#instance.

Also, since we are making changes to [markdown.js](https://github.com/Adamant-im/adamant-im/blob/dev/src/lib/markdown.js#L1-L19), we can convert it to TypeScript.

skranee:
[Code: Upgrade marked 14 → 15](https://trello.com/c/KInLQ6sa/681-code-upgrade-marked-14-%E2%86%92-15)

src/lib/markdown.ts
```diff
@@ -1,28 +1,42 @@
...
+
 // The U+2028 character (LINE SEPARATOR) is sometimes used as a line break, but it is treated as a space in some web environments,
 // causing unexpected rendering issues. To avoid this, it's recommended to replace it with a standard line break character such as `\n`.
-const LINE_SEPARATOR = /\u2028/g
+const LINE_SEPARATOR: RegExp = /\u2028/g
 
 // LINE_BREAK_VISUAL is used for formatting when the message is displayed as a single line in a preview.
 // It replaces line breaks with a visual symbol (↵) to indicate where new lines exist in the original text.
-const LINE_BREAK_VISUAL = '↵ '
+const LINE_BREAK_VISUAL: string = '↵ '
```
bludnic, line 22: You can omit typing here, because TS can infer the type from the actual value. Check other places as well.
https://www.typescriptlang.org/docs/handbook/type-inference.html

src/lib/markdown.ts
```diff
@@ -1,28 +1,42 @@
...
   // sanitize: true,
   gfm: true,
   breaks: true
 })
 
-const renderer = new marked.Renderer()
+const renderer = new markedInstance.Renderer()
 
-renderer.image = function ({ _href, _title, _text }) {
+renderer.image = function ({ _href, _title, _text }: IImageParams): string {
```
bludnic, line 35: This may also work with Type Inference. You don't need to define custom types (IImageParams, ILinkParams).

src/lib/markdown.ts
```diff
@@ -9,20 +9,22 @@ const LINE_SEPARATOR = /\u2028/g
 // It replaces line breaks with a visual symbol (↵) to indicate where new lines exist in the original text.
 const LINE_BREAK_VISUAL = '↵ '
 
-marked.setOptions({
+const markedInstance = new Marked()
```
bludnic, line 12: Suggestion: This is also fine. Resolve if you leave as is.

```suggestion
const marked = new Marked()
```

src/lib/markdown.ts
```diff
@@ -65,19 +67,19 @@ export function renderMarkdown(text = '') {
  * @param {string} text text to process
  * @returns {string} resulting clear text of the first line
  */
-export function removeFormats(text = '') {
+export function removeFormats(text = ''): string | Promise<string> {
```
bludnic, line 70: Return type can also be inferred.

src/lib/markdown.ts
```diff
@@ -65,19 +67,19 @@ export function renderMarkdown(text = '') {
...
   const node = document.createElement('div')
   const textWithSymbol = text.replace(/\n/g, '↵ ')
-  node.innerHTML = marked.parse(sanitizeHTML(textWithSymbol))
+  node.innerHTML = markedInstance.parse(sanitizeHTML(textWithSymbol)) as string
 
   return node.textContent || node.innerText || ''
 }
 
-export function formatMessage(text = '') {
+export function formatMessage(text = ''): string | Promise<string> {
```
bludnic, line 78: Same. Try inferred return type

bludnic (changes requested):

src/lib/markdown.ts
bludnic: After adding `{ async: false }` https://github.com/Adamant-im/adamant-im/pull/730#discussion_r1971989526, it is important to check that the function does not return a Promise, otherwise it will break the logic in Vue components.

skranee: When I tried to rely on typescriipt here, it showed me return type 'string' and that caused errors while building the app, that's why I typed that.

src/lib/markdown.ts
skranee: Same as I replied above.

skranee: I guess you were talking about omitting typing after passing { async: false }, my bad

bludnic (approved):

bludnic:
Looks good to me 👍

S-FrontendDev (approved):

PR #736, opened 2025-03-03
refactor: replace qs lib with native URLSearchParams

skranee:
https://trello.com/c/QwbYMolz/680-code-replace-qs-dependency-with-native-urlsearchparams


S-FrontendDev (approved):

src/lib/bitcoin/doge-api.js
```diff
@@ -157,8 +156,11 @@ export default class DogeApi extends BtcBaseApi {
 
   /** Executes a POST request to the DOGE API */
   _post(url, data) {
+    const flatData = flattenData(data);
```
bludnic, line 159: Can you explain why this is needed? Preferably with examples.

bludnic: Seems overengineered. I'd prefer to use native URLSearchParams (or alternative) functionality, instead of a custom function.

bludnic (changes requested):

src/lib/bitcoin/doge-api.js
skranee: This is needed because native URLSearchParams can not work with deep objects, it can only process flat ones. So if the the 'data' object is going to have some depth, the params would look like that
image

skranee: In qs lib it's processed by default up to object's depth = 5, but while using native URLSearchParams you should either be sure that the object is always flat and has no depth or flatten it on your own to avoid results as is shown on the screenshot.

skranee: Also, if I'm not mistaken, a native flat() method is only available for arrays in JS. If we're working with object, we have to flatten the object using your own functions (that's also what I've got while searching the answer on stack overflow and through AI requests).

bludnic: I get your point, but does the BTC, DASH or DOGE APIs use nested params? Perhaps we don't this at all.

skranee: IDK about that honestly, will do a research later

skranee:
URLSearchParams turned out to be redundant, so I removed it. Everything works as expected with just passing the object

bludnic (approved):

PR #742, opened 2025-03-13
feat: chats one screen with swipe

S-FrontendDev:
if swipe is not good i will create new PR without it

### DIFF
--- a/src/hooks/useSwipeRight.ts
+++ b/src/hooks/useSwipeRight.ts
@@ -1,0 +1,59 @@
+import { ref } from 'vue'
+
+/**
+ * Ratio between swipe offsetX / offsetY.
+ * The higher value, the smoother swipe is needed on X axis
+ * to activate the swipe event.
+ * @type {number}
+ */
+const SWIPE_RATIO_ACTIVATION = 2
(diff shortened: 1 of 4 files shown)


S-FrontendDev:
[Feat: UX. Contact list and chat messages on the same screen (for Desktop)](https://trello.com/c/eYF403LQ/644-feat-ux-contact-list-and-chat-messages-on-the-same-screen-for-desktop)

src/hooks/useSwipeRight.ts
```diff
@@ -0,0 +1,60 @@
+import { ref } from 'vue'
+import { VuetifyTouchEvent } from '@/types/vuetify'
+
+/**
+ * Ratio between swipe offsetX / offsetY.
+ * The higher value, the smoother swipe is needed on X axis
+ * to activate the swipe event.
+ * @type {number}
+ */
+const SWIPE_RATIO_ACTIVATION = 2
```
skranee, line 10: May be it'd be a good idea to keep all constants in a specialized file in case there'll be more swiper in the app.

skranee (approved):

PR #746, opened 2025-03-21
Feat/spinner when restore

skranee:
https://trello.com/c/xkUCTzhD/590-connection-feat-show-loading-spinner-when-restore-from-background-copied-from-ios

I've added loaders in some places where they'd be seen and not look awful (imo). If needed to be put in different places - suggest please


skranee:
[{Connection} Feat: Show loading spinner when restore from background (copied from iOS)](https://trello.com/c/xkUCTzhD/590-connection-feat-show-loading-spinner-when-restore-from-background-copied-from-ios)

src/views/Chats.vue
```diff
@@ -97,10 +105,13 @@ export default {
   data: () => ({
     showChatStartDialog: false,
     loading: false,
-    noMoreChats: false
+    noMoreChats: false,
   }),
   computed: {
     className: () => 'chats-view',
+    enabledNodes() {
```
adamantmm, line 112: Criteria for spinner in Chats:
All chats are up to date.
It means that a user is sure that they see actual chats.
You can ask iOS team how they did it.

src/components/Chat/Chat.vue
```diff
@@ -74,11 +74,21 @@
         <chat-toolbar :partner-id="partnerId">
           <template #avatar-toolbar>
             <ChatAvatar
+              v-if="enabledNodes"
               class="chat-avatar"
               :user-id="partnerId"
               use-public-key
               @click="onClickAvatar(partnerId)"
             />
+            <v-progress-circular
```
adamantmm, line 83: Criteria for spinner inside a Chat:
All messages are fetched and up to date.
It means that a user is sure that they see actual message.
You can ask iOS team how they did it.

adamantmm (changes requested):

src/views/Chats.vue
skranee: IOS team told they've done it the same way as I did. Just watch for nodes status and when they're up again -> loader disappears and chats are considered up to date

src/components/Chat/Chat.vue
skranee: Replied above

src/components/Chat/Chat.vue
```diff
@@ -74,11 +74,21 @@
...
+              v-else
+              :class="`mr-3`"
```
bludnic, line 85: There is no need in string literal here

```suggestion
              class="mr-3"
```

src/components/Chat/Chat.vue
```diff
@@ -74,11 +74,21 @@
...
+              v-show="!enabledNodes"
+              indeterminate
+              color="secondary"
+              size="30"
+              style="z-index: 100"
```
bludnic, line 90: Can you explain why ZIndex here? Does it work without?

src/components/Chat/Chat.vue
```diff
@@ -74,11 +74,21 @@
...
               use-public-key
               @click="onClickAvatar(partnerId)"
             />
+            <v-progress-circular
+              v-else
+              :class="`mr-3`"
+              v-show="!enabledNodes"
+              indeterminate
+              color="secondary"
+              size="30"
```
bludnic, line 89: Better to pass as a `number`

```suggestion
              :size="30"
```

bludnic: All the numbers (paddings, margins, width, height) in design system must be dibisible by 4 or 8. Better to use `32`.

src/components/Chat/Chat.vue
```diff
@@ -353,6 +364,8 @@ const showEmojiPicker = ref(false)
 
 const messages = computed(() => store.getters['chat/messages'](props.partnerId))
 const userId = computed(() => store.state.address)
+const enabledNodes = computed(() => store.getters['nodes/adm']
+  .filter((node: NodeStatusResult) => node.status === 'online').length)
```
bludnic, line 368:
```suggestion
const admNodes = computed<NodeStatusResult[]>(() => store.getters['nodes/adm'])
const admNodesOnline = computed(() => admNodes.value.some((node) => node.status === 'online'))
```

src/components/nodes/NodesTable.vue
```diff
@@ -1,6 +1,14 @@
 <template>
   <div :class="classes.root">
     <v-tabs v-model="tab" bg-color="transparent">
+      <v-progress-circular
+        :class="`align-self-center ml-3 mr-2`"
```
bludnic, line 5: No need in string literal

bludnic (changes requested):

src/components/Chat/Chat.vue
skranee: Overlooked that, was copied from another file

src/components/Chat/Chat.vue
```diff
@@ -74,11 +74,20 @@
...
               :user-id="partnerId"
               use-public-key
               @click="onClickAvatar(partnerId)"
             />
+            <v-progress-circular
+              v-else
+              class="mr-3"
+              v-show="!admNodesOnline || !chatActual"
+              indeterminate
+              color="secondary"
```
adamantmm, line 88: Everywhere for loading spinner

Dark theme: 'regular': #4a4a4a
Light theme: 'grey': #9e9e9e

src/components/Chat/Chat.vue
```diff
@@ -74,11 +74,20 @@
         <chat-toolbar :partner-id="partnerId">
           <template #avatar-toolbar>
             <ChatAvatar
+              v-if="admNodesOnline && chatActual"
               class="chat-avatar"
               :user-id="partnerId"
               use-public-key
               @click="onClickAvatar(partnerId)"
             />
+            <v-progress-circular
```
adamantmm, line 83: Add paddings so that when spinner/avatar change doesn't jerk.

src/components/Chat/Chat.vue
```diff
@@ -422,6 +444,13 @@ onBeforeMount(() => {
   window.addEventListener('keyup', onKeyPress)
 })
 onMounted(() => {
+  // To show the spinner if chats are not actual anymore
+  chatActualInterval.value = setInterval(() => {
```
adamantmm, line 448: Why do we use `setInterval()` here?
Why can't we watch `chatsActualUntil`?

src/components/Chat/Chat.vue
```diff
@@ -350,9 +360,14 @@ const actionsMenuMessageId = ref<string | -1>(-1)
 const actionsDropdownMessageId = ref<string | -1>(-1)
 const replyMessageId = ref<string | -1>(-1)
 const showEmojiPicker = ref(false)
+const chatActualInterval = ref<ReturnType<typeof setInterval> | undefined>(undefined)
 
 const messages = computed(() => store.getters['chat/messages'](props.partnerId))
 const userId = computed(() => store.state.address)
+const admNodes = computed<NodeStatusResult[]>(() => store.getters['nodes/adm'])
+const admNodesOnline = computed(() => admNodes.value.some((node) => node.status === 'online'))
+const chatActual = computed(() => store.state.chat.chatsActual)
```
adamantmm, line 369: Why is it reasonable here?

src/components/Chat/Chat.vue
```diff
@@ -395,6 +410,13 @@ watch(
   }
 )
 
+// To update immediately if the message was sent using sockets
+watch(chatsActualUntil, () => {
+  const areChatsActual = chatsActualUntil.value > Date.now()
+
+  store.commit('chat/setChatsActual', areChatsActual)
```
adamantmm, line 417: Why do we store something here?

src/components/Chat/Chat.vue
```diff
@@ -395,6 +410,13 @@ watch(
   }
 )
 
+// To update immediately if the message was sent using sockets
+watch(chatsActualUntil, () => {
+  const areChatsActual = chatsActualUntil.value > Date.now()
```
adamantmm, line 415:
```suggestion
  const chatActual = chatsActualUntil.value > Date.now()
```

src/store/modules/chat/index.js
```diff
@@ -51,7 +51,9 @@ const state = () => ({
   lastMessageHeight: 0, // `height` value of the last message
   isFulfilled: false, // false - getChats did not start or in progress, true - getChats finished
   offset: 0, // for loading chat list with pagination. -1 if all of chats loaded
-  noActiveNodesDialog: undefined // true - visible dialog, false - hidden dialog, but shown before, undefined - not shown
+  noActiveNodesDialog: undefined, // true - visible dialog, false - hidden dialog, but shown before, undefined - not shown
+  chatsActual: false,
```
adamantmm, line 55: It's excessive to store `chatsActual` as we count it for each screen.
`chatsActualUntil` is enough.

src/store/modules/chat/index.js
```diff
@@ -76,6 +78,15 @@ const getters = {
     return []
   },
 
+  /**
+   * Returns the timeout for chatsActual
+   * @depends options.useSocketConnection
+   * @returns {number}
+   */
+  chatsActualTimeout: (state, getters, rootState) => {
```
adamantmm, line 86: Excessive as we update `chatsActualUntil` when new messages arrive whenever it will be.

Do we store it somewhere already?

src/store/modules/chat/index.js
```diff
@@ -358,6 +369,14 @@ const mutations = {
     state.isFulfilled = value
   },
 
+  setChatsActual(state, value) {
```
adamantmm, line 372: Excessive

src/store/plugins/socketsPlugin.js
```diff
@@ -17,6 +17,12 @@ function subscribe(store) {
       // So we'll update confirmations in getTransactionStatus()
 
       if (transaction.asset?.chat?.type !== MessageType.SIGNAL_MESSAGE) {
+        const timestamp = transaction.timestamp
+        const chatsActualTimeout = store.getters['chat/chatsActualTimeout']
+
+        const validUntil = (timestamp * 1000) + chatsActualTimeout + EPOCH + 1000
+
+        store.commit('chat/setChatsActualUntil', validUntil)
```
adamantmm, line 25: It's the main part.

src/views/Chats.vue
```diff
@@ -13,6 +13,13 @@
                 size="small"
                 variant="text"
               />
+              <v-progress-circular
```
adamantmm, line 16: Same logic here as for Chat.vue

adamantmm (changes requested):

adamant-al (changes requested):

src/components/AppToolbarCentered.vue
```diff
@@ -93,6 +106,13 @@ export default {
     letter-spacing: 0.02em;
   }
 
+  .spinner {
```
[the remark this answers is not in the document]
skranee: But you said that spinner must be after the text instead of the center on devices with little width (I thought it means for mobile devices)

src/store/modules/chat/index.js
```diff
@@ -645,16 +665,20 @@ const actions = {
...
     return getChats(state.lastMessageHeight).then((result) => {
-      const { messages, lastMessageHeight } = result
+      const { messages, lastMessageHeight, nodeTimestamp } = result
+      const chatsActualInterval = getters.chatsPollingTimeout
 
       dispatch('pushMessages', messages)
 
+      const validUntil = nodeTimestamp * 1000 + chatsActualInterval + EPOCH + BUFFER_TIMEOUT
+
+      commit('setChatsActualUntil', validUntil)
```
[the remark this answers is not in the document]
skranee: Not sure I understand what's the change here

src/views/Chats.vue
```diff
@@ -72,7 +78,8 @@ import NodesOfflineDialog from '@/components/NodesOfflineDialog.vue'
 import scrollPosition from '@/mixins/scrollPosition'
 import { getAdamantChatMeta, isAdamantChat, isStaticChat } from '@/lib/chat/meta/utils'
 import { mdiMessageOutline, mdiCheckAll } from '@mdi/js'
-
+import { useIntervalFn } from '@vueuse/core'
```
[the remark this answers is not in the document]
skranee: `useNow` is easier to use in Composition API, but in Options API `useNow` is not that convenient so I decided to use `useIntervalFn` (`useNow` also uses `useIntervalFn`)

src/components/AppToolbarCentered.vue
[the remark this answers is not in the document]
adamant-al: It remains always centered on all devices, while there is space in the center. Not to override text.

src/store/plugins/socketsPlugin.js
```diff
@@ -17,6 +17,12 @@ function subscribe(store) {
       // So we'll update confirmations in getTransactionStatus()
 
       if (transaction.asset?.chat?.type !== MessageType.SIGNAL_MESSAGE) {
+        const timestamp = transaction.timestamp
+        const chatsActualTimeout = store.getters['chat/chatsPollingTimeout']
+
+        const validUntil = (timestamp * 1000) + chatsActualTimeout + EPOCH + BUFFER_TIMEOUT
```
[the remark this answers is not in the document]
adamant-al: You are right.
If we're already in the chatActual status, 
we extend the validity of the status by chatsActualTimeout seconds,
**BUT limit by Now() + chatsActualTimeout**.

Don't use BUFFER_TIMEOUT for sockets.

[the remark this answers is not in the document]
skranee: But why do we need to extend the validity time if we can just replace it with a new one? Imo it's a better option because it represents the logic behind being sure that the user has fresh messages. I don't think it should depend on the way of getting new messages.
1. Got an answer from 'fromHeight' request? => Good, now the user has the data that is considered the newest until certain time.
2. Got a message through sockets? => Good, now we have the latest messages that we got anyways.
Other way we might face next problem:
1. We get an answer from a 'fromHeight' request => +11 seconds to validity.
2. We get a message through sockets => +10 seconds to validity.
Why should we have 21 seconds of validity instead of 10-11?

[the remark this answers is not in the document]
adamant-al: > Got a message through sockets? => Good, now we have the latest messages that we got anyways.

Unfortunately, no.
We can receive a new message via socket, but last 100 messages are yet missed.
Maybe it's better remove `chatActualUntil` logic from socket at all and use REST only.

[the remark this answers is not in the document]
skranee: That sounds rational. I guess you're right and we must rely on REST only in case we temporarily lose socket subscription. Agree?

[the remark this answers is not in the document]
adamant-al: We can simplify and rely on REST in any case.

As an alternative for socket,

If we're already in the chatActual status,
we extend the validity of the status by chatsActualTimeout seconds,
BUT limit by Now() + chatsActualTimeout.

This can do connection spinner more stable.

adamant-al (approved):
👍

src/views/Chats.vue
```diff
@@ -97,10 +104,22 @@ export default {
...
     className: () => 'chats-view',
+    admNodes() {
+      return this.$store.getters['nodes/adm']
+    },
+    admNodesOnline() {
+      return this.admNodes.some((node) => node.status === 'online')
+    },
+    chatsActualUntil() {
+      return this.$store.state.chat.chatsActualUntil
+    },
```
S-FrontendDev, line 122: also to setup

S-FrontendDev: maybe we have status in enum or const?

src/views/Chats.vue
```diff
@@ -139,11 +159,35 @@ export default {
...
+    // Hides loading spinner once the latest messages are fetched
+    chatsActualUntil: {
+      handler: 'updateChatsActual'
+    },
+    // Shows loading spinner when chats become outdated
+    currentTime: {
+      handler: 'updateChatsActual',
+      immediate: true
+    }
   },
```
S-FrontendDev, line 191: setup

src/views/Votes.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div :class="className">
-    <app-toolbar-centered app :title="t('votes.page_title')" flat fixed />
+    <app-toolbar-centered app :title="t('votes.page_title')" :has-spinner="true" flat fixed />
```
S-FrontendDev, line 3: just has-spinner

src/components/Chat/Chat.vue
```diff
@@ -428,6 +459,12 @@ onMounted(() => {
     isScrolledToBottom.value = chatRef.value.isScrolledToBottom()
   })
   visibilityId.value = Visibility.change((event, state) => {
+    if (state === 'visible') {
+      resume()
```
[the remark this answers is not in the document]
skranee: ? we need to resume the timer if the visibilite changes to 'visible', it doesn't start on its own

src/views/Chats.vue
```diff
@@ -97,10 +104,22 @@ export default {
   data: () => ({
     showChatStartDialog: false,
     loading: false,
-    noMoreChats: false
+    noMoreChats: false,
+    chatsActual: false,
+    currentTime: Date.now(),
+    visibilityId: null
```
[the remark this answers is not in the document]
skranee: I don't see the point of mixing options API and composition API that badly. I understand if it's already mixed, but when we have almost empty setup() I consider it's a bad practice to put some computed in setup() and leave some of them in options API. That's what official vue documentation thinks of it - https://vuejs.org/guide/extras/composition-api-faq.html#can-i-use-both-apis-in-the-same-component.  I totally agree with vue guideline on that and don't see the point of not using all benefits of composition API while also ruining the structure of options API.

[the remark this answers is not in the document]
S-FrontendDev: I absolutely do not see a recommendation not to mix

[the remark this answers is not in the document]
S-FrontendDev: there is no point in writing new code on the old API

src/components/Chat/Chat.vue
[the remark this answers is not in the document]
skranee: can I resolve that? I don't get the purpose of the change, it'll break the logic

bludnic (changes requested):

src/hooks/useChatsSpinner.ts
```diff
@@ -0,0 +1,38 @@
...
+  const visibilityId = ref<number | boolean | null>(null)
+
+  const admNodes = computed<NodeStatusResult[]>(() => store.getters['nodes/adm'])
+  const admNodesOnline = computed(() => admNodes.value.some((node) => node.status === 'online'))
+  const chatsActualUntil = computed(() => store.state.chat.chatsActualUntil)
+  const currentTime = computed(() => now.value.getTime())
+
+  watchEffect(() => {
+    chatsActual.value = chatsActualUntil.value > currentTime.value
+  })
```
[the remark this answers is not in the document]
skranee: Makes sense

adamantmm (approved):

adamantmm:
@bludnic 
Please review

adamant-al:
@skranee 
Please update.
Merge candidate.

src/components/Chat/Chats.vue
```diff
@@ -67,268 +73,244 @@
...
-      }
+const attachScrollListener = () => {
+  window.addEventListener('scroll', onScroll)
+}
+
+const destroyScrollListener = () => {
+  window.removeEventListener('scroll', onScroll)
+}
+
+const onScroll = (event?: Event) => {
```
bludnic, line 251: Is there a scenario when event is `undefined`? @skranee @S-FrontendDev

src/components/Chat/Chats.vue
```diff
@@ -67,268 +73,244 @@
...
+const loading = ref(false)
+const noMoreChats = ref(false)
+const loadingSeparator = ref<InstanceType<typeof ChatPreview>[]>([])
+
+const chatPagePartnerId = computed(() => {
+  const partnerId = route.params.partnerId
+
+  // To convert it to string immediately => no need in further type conversion
+  return Array.isArray(partnerId) ? partnerId[0] : partnerId
+})
```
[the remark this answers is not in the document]
skranee: If we're 100% sure about it than yes, `as` would be acceptable in that case then I think

src/components/Chat/Chats.vue
```diff
@@ -67,268 +73,244 @@
...
+      if (!route.query.from) {
+        return router.push({ name: 'Chats' })
       }
 
-      if (!value && route.name === 'Chats') {
-        savedRoute.value = null
-      }
-    })
+      const from = Array.isArray(route.query.from) ? route.query.from[0] : route.query.from
+      return from ? router.push(from) : router.push({ name: 'Chats' })
```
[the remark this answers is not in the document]
skranee: Same as above

src/components/Chat/Chats.vue
skranee: image

bludnic (approved):

PR #757, opened 2025-04-05
Add workflow for CodeQL scan

graycraft:


graycraft:
[Code: Security code scanner](https://trello.com/c/6Bkssk8h/701-code-security-code-scanner)

skranee (approved):

.github/workflows/codeql-scan.yml
```diff
@@ -0,0 +1,54 @@
+name: CodeQL Scan
```
bludnic, line 1: Does it scan only our source code? What about scanning `node_modules`, in the `/dist` dir after build? Did you check how Metamask is dealing with that?

graycraft: Metamask uses own custom scanner, `dist` skipped because basically it has the same logic as `src` directory.
`node_modules` can be easily scanned by [`npm audit`](https://docs.npmjs.com/cli/v11/commands/npm-audit)

bludnic: We can't rely only on `npm audit`. If no one reported a malicious code inside the package, the NPM audit will not catch it.

I think we need to scan it additionally with CodeQL, either the `node_modules` or the `dist` dir.

bludnic: Also do a research about other open source projects (besides MetaMask), how they a dealing with that? @graycraft

adamant-al: CodeQL: CodeQL can be configured to scan node_modules as well. This can help identify potential security issues that npm audit might miss. You can modify your workflow to include node_modules in the scan.

Suggested additions:

- Add npm audit or even audit-ci in CI as a separate job.
- Use a static analysis tool like [Snyk](https://snyk.io/) or [Socket.dev](https://socket.dev/) to catch malicious patterns (not just vulnerabilities).
- Don’t scan dist — that’s compiled code. Instead, ensure your source code and build pipeline are secure.

Here are some examples you could explore:

- MetaMask: custom pipeline + not relying on dist/node_modules.
- Hardhat (Nomic Foundation): uses npm audit + some static analysis.
- Ethers.js: relies on lightweight CI and minimal deps.
- Zcash: uses CodeQL + other tools to scan for logic issues.
- Brave Browser: runs both CodeQL and custom static scans with alerts for dependency changes.
- Projects like Webpack, Next.js, and other popular JavaScript/TypeScript projects might have robust security practices that you can learn from.

TL;DR Suggestions:

- Use CodeQL for source code.
- Use audit-ci or snyk test to scan node_modules.
- Don't scan dist, it's unnecessary duplication.
- Consider adding npm audit --audit-level=high as a gate.
- Review tools like Snyk, Socket.dev, or NodeSecure (https://github.com/nodesecure/nodesecure) for additional layer of scanning.

PR #759, opened 2025-04-08
fix: qr code - read from canvas

S-FrontendDev:


src/components/QrcodeCapture.vue
```diff
@@ -13,18 +13,56 @@
...
 <script>
 import { vibrate } from '@/lib/vibrate'
+import { useTemplateRef } from 'vue'
 
 export default {
   emits: ['detect', 'error'],
+  setup() {
+    const canvas = useTemplateRef('canvasElement')
+
+    const IMG_MAX_SIZE = 400
```
bludnic, line 29: Move outside Vue component

src/components/QrcodeCapture.vue
```diff
@@ -13,18 +13,56 @@
       @change="onFileSelect"
     />
 
-    <img ref="imageElement" :src="imageBase64" :class="`${className}__image`" />
+    <canvas ref="canvasElement" :class="`${className}__image`" />
   </div>
 </template>
 
 <script>
```
bludnic, line 20: Can you convert it fully to Composition API, and also use TS?

src/components/QrcodeCapture.vue
```diff
@@ -13,18 +13,56 @@
...
+          canvas.value.height = newHeight
+
+          ctx.drawImage(img, 0, 0, canvas.value.width, canvas.value.height)
+
+          resolve(canvas)
+        }
+
+        img.onerror = (e) => reject(e)
+
+        img.src = URL.createObjectURL(file)
```
bludnic, line 57: At some point you should call [URL.revokeObjectURL()](https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL_static), otherwise the object will still exist even after user logsout.

I think you can revoke it right after the promise was resolved.

src/components/QrcodeCapture.vue
bludnic: I meant above the Vue component definition, not in another file. If you converted to script setup, it's fine.

src/components/QrcodeCapture/QrcodeCapture.vue
```diff
@@ -0,0 +1,153 @@
...
+
+      revokeObjectURL()
+
+      resolve(canvas)
+    }
+
+    img.onerror = (e) => reject(e)
+
+    img.src = imgUrl
+  })
```
bludnic, line 83: Instead of invoking `URL.revokeObjectURL()` in several places you can do it once using by `.finally()`. Example:

```js
const promise = new Promise(...)
promise.finally(() => URL.revokeObjectURL(imgUrl))

return promise
```

src/components/QrcodeCapture/QrcodeCapture.vue
```diff
@@ -0,0 +1,153 @@
+<template>
+  <div :class="className">
+    <div :class="`${className}__activator`" @click="fileInput?.click()">
```
bludnic, line 3: Check the Template classes rules https://developers.adamant.im/template-classes.html

src/components/QrcodeCapture/QrcodeCapture.vue
```diff
@@ -0,0 +1,153 @@
...
+
+    <canvas ref="canvasElement" :class="`${className}__image`" />
+  </div>
+</template>
+
+<script setup lang="ts">
+import { vibrate } from '@/lib/vibrate'
+import { useTemplateRef } from 'vue'
+import { IMG_MAX_SIZE } from '@/components/QrcodeCapture/consts'
+import type { BrowserQRCodeReader } from '@zxing/browser/esm/readers/BrowserQRCodeReader'
```
bludnic, line 24:
```suggestion
import type { BrowserQRCodeReader } from '@zxing/browser'
```

src/components/QrcodeCapture/QrcodeCapture.vue
```diff
@@ -0,0 +1,153 @@
...
+
+const emit = defineEmits<{
+  (e: 'detect', text: string): void
+  (e: 'error', err: unknown): void
+}>()
+
+const className = 'qrcode-capture'
+
+let qrCodeText = ''
+let codeReader: BrowserQRCodeReader
```
bludnic, line 34: Why do you store these variables without `ref()`?

src/components/QrcodeCapture/QrcodeCapture.vue
```diff
@@ -0,0 +1,153 @@
...
+    img.src = imgUrl
+  })
+}
+
+/**
+ * Decode QRCode from canvas.
+ * @returns {Promise<string>}
+ */
+const getQrcode = async () => {
+  const result = await codeReader?.decodeFromCanvas(canvas.value as HTMLCanvasElement)
```
bludnic, line 91: Remove `await`, it does not return a Promise.

```suggestion
  const result = codeReader?.decodeFromCanvas(canvas.value as HTMLCanvasElement)
```


If you want to tell TS that this value is required at this stage, use **Non-null assertion operator** instead, but with caution. I don't recommend using this pattern a lot.

```suggestion
  const result = await codeReader?.decodeFromCanvas(canvas.value!)
```

src/components/QrcodeCapture/QrcodeCapture.vue
```diff
@@ -0,0 +1,153 @@
...
+  return new Promise<string>((resolve, reject) => {
+    // Vue should rerender <img> element,
+    // so add a callback to the macrotasks queue
+    setTimeout(() => {
+      getQrcode()
+        .then((qrCodeText) => resolve(qrCodeText))
+        .catch((err) => reject(err))
+    }, 0)
+  })
+}
```
bludnic, line 106: `getQrcode()` does not return a promise anymore, so you probably can just remove the `tryToDecode` function

src/components/QrcodeCapture/QrcodeCapture.vue
```diff
@@ -0,0 +1,153 @@
...
+      codeReader = new BrowserQRCodeReader()
+    }
+
+    const { target } = event
+
+    if (!target) {
+      throw Error('no target')
+    }
+
+    const file = (target as HTMLInputElement)?.files?.[0]
```
bludnic, line 121: I suggest you to cast the type as earlier as possible. Also you probably don't need to check the `target` on `null` (it is always defined). The code looks cleaner.

```suggestion
    const input = event.target as HTMLInputElement;
    const file = input.files?.[0]
```

src/components/QrcodeCapture/QrcodeCapture.vue
```diff
@@ -0,0 +1,153 @@
...
+
+    if (!file) {
+      throw Error('no file')
+    }
```
bludnic, line 125: How this will work from the user perspective? A snackbar with error will be shown? And can you reproduce that scenario?

bludnic (changes requested):

src/components/QrcodeCapture/QrcodeCapture.vue
S-FrontendDev: you asked just about rewrite from options api, it's not my component. So i didn't refact it. Just move to folder to create const's

src/components/QrcodeCapture/QrcodeCapture.vue
S-FrontendDev: it's not reactive vars

S-FrontendDev: just let enougth

src/components/QrcodeCapture/QrcodeCapture.vue
S-FrontendDev: idk how to reproduce, but TS us file can be undefined

so if file is undefined then we will get snackbar with "“QR code does not contain passphrase or is unrecognizable”"

src/components/QrcodeCapture/QrcodeCapture.vue
bludnic: The `<QrcodeCapture />` component is reused in multiple instances, while `qrCodeText` and `codeReader` vars are common. This may lead to unpredictable behavior.

bludnic: Can you share the link to Vue docs where this approach is used? @S-FrontendDev

src/components/QrcodeCapture/QrcodeCapture.vue
bludnic: I get it. Since we refactored the component, fixing the classes seemed like an easy change.

src/components/QrcodeCapture/QrcodeCapture.vue
bludnic: Let's use the non-null assertion operator to simplify the code. If an error occurs, it will be caught anyway with the same behavior.
```js
const file = input.files?.[0]!
```

src/components/QrcodeCapture.vue
```diff
@@ -9,88 +9,141 @@
...
+    if (!file) {
+      throw Error('no file')
+    }
+
+    await drawCanvas(file)
+    qrCodeText = await tryToDecode()
+
+    vibrate.veryShort()
+
+    emit('detect', qrCodeText)
```
bludnic, line 137: Intermediate `let qrCodeText` at the top of component probably not required, just remove it

bludnic (changes requested):

src/components/QrcodeCapture.vue
```diff
@@ -9,88 +9,141 @@
...
+  return promise
+}
 
-        this.$emit('detect', this.qrCodeText)
-      } catch (err) {
-        vibrate.tripleVeryShort()
+/**
+ * Decode QRCode from canvas.
+ * @returns {<string>}
+ */
```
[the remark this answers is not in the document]
S-FrontendDev: jsDoc not only for types, it can discrabe props, methods and etc

src/components/QrcodeCapture/QrcodeCapture.vue
S-FrontendDev: ?? 

i'm waiting for setTimeout. And i need resolve promise. Because we need render canvas

src/components/QrcodeCapture/QrcodeCapture.vue
S-FrontendDev: need to add eslint ignore. Are you sure?

src/components/QrcodeCapture.vue
```diff
@@ -9,88 +9,132 @@
...
+  try {
+    if (!codeReader) {
+      const { BrowserQRCodeReader } = await import('@zxing/browser')
+      codeReader = new BrowserQRCodeReader()
     }
+
+    const input = event.target as HTMLInputElement
+
+    // eslint-disable-next-line @typescript-eslint/no-non-null-asserted-optional-chain
+    const file = input.files?.[0]!
```
skranee, line 121: Will it not be better to manually check ```if (!file) return```?

skranee (changes requested):

bludnic (approved):

src/components/QrcodeCapture/QrcodeCapture.vue
bludnic: Approved in advance @S-FrontendDev

bludnic (approved):

PR #761, opened 2025-04-10
fix: pending tx fee

S-FrontendDev:


S-FrontendDev:
[Bug: Wrong tx fee for pending txn](https://trello.com/c/ZJGD1bPc/751-bug-wrong-tx-fee-for-pending-txn)

adamantmm (approved):

src/components/transactions/TransactionTemplate.vue
```diff
@@ -283,43 +283,41 @@ export default defineComponent({
...
 
       const { cryptoTransferDecimals, decimals } = CryptosInfo[commissionTokenLabel]
 
-      const tokenFee = typeof props.fee === 'number'
-        ? `${formatAmount(props.fee, cryptoTransferDecimals ?? decimals)} ${commissionTokenLabel}`
-        : placeholder.value;
-
-      if (!props.fee || !calculatedTimestampInSec.value) return tokenFee;
+      const tokenFee =
+        (props.queryStatus === 'success' && typeof props.fee === 'number') || props.fee
```
bludnic, line 305: Does removing `|| props.fee` change the behavior, or will it work the same?

S-FrontendDev: typeof props.fee === 'number' true when props.fee === 0, idk may be fee can be 0 for some cryptos and we will show it when we get queryStatus === 'success' . But if we got fee early we will show it

bludnic: So `|| props.fee` doesn't affect the logic, and we can remove it?

bludnic (approved):

PR #763, opened 2025-04-12
fix: electron build esm

S-FrontendDev:
type module broken electron, fix for esm


S-FrontendDev:
[Fix: electron build broken](https://trello.com/c/nhoweUgq/762-fix-electron-build-broken)

src/electron/main.js
```diff
@@ -1,10 +1,12 @@
 import { app, BrowserWindow, Menu, nativeTheme, protocol } from 'electron'
+import { fileURLToPath, URL } from 'url'
```
bludnic, line 2: I suggest to prefix NodeJS core modules with `node:`, to not confuse with packages from `node_modules`. Same for `path`, `fs`, etc.

```suggestion
import { fileURLToPath, URL } from 'node:url'
```

vite-base.config.ts
```diff
@@ -15,6 +16,9 @@ import { excludeBip39Wordlists } from './vite-config/rollup/excludeBip39Wordlist
 const env = loadEnv('production', process.cwd())
 const basePublicPath = env.VITE_PUBLIC_PATH || '/'
 
+const __filename = fileURLToPath(import.meta.url)
+const __dirname = path.dirname(__filename)
```
bludnic, line 20: I don't see you use `__filename` or `__dirname` in that file.

bludnic (changes requested):

bludnic:
@S-FrontendDev Have you checked the docs on using Electron with ESM? https://www.electronjs.org/docs/latest/tutorial/esm

vite-base.config.ts
bludnic: I missed it. Everything is fine

S-FrontendDev:
> @S-FrontendDev Have you checked the docs on using Electron with ESM? https://www.electronjs.org/docs/latest/tutorial/esm

I think my code follows the documentation

PR #764, opened 2025-04-13
fix: remove autofocus on field

skranee:


skranee:
[Bug/UI: Login by password (not passphrase) field don't fit 100% width iOS device](https://trello.com/c/jMMZOxBB/757-bug-ui-login-by-password-not-passphrase-field-dont-fit-100-width-ios-device)

src/components/LoginPasswordForm.vue
```diff
@@ -4,7 +4,6 @@
       <v-text-field
         ref="passwordInput"
         v-model="password"
-        autofocus
```
bludnic, line 7: Does removing `autofocus` fix the bug? What if we want to keep autofocus behaviour?

skranee: Yes it does, if we wanna keep the behavior, I guess we must either set it properly or find another way of auto focusing. If I got it right, previous autofocus focused not the center of the field.

skranee: Also it might be something wrong with iOS. Need to test it on android device probably

skranee: Also it doesn't really work as autofocus. When reloading the page it doesn't focus the input field, it just zooms it after manually focusing.

adamantmm: > When reloading the page it doesn't focus the input field
It does. Login by password, not passphrase.
Is there other way to resolve the issue?

skranee: > It does.
Yeah, my bad. On the phone it actually focuses the field, but there's no cursor and keyboard so I thought it doesn't focus, I'll try to figure all of that out to accomplish autofocusing on the center of the field while logging in with password and also avoid redundant moving of the label.

src/components/LoginPasswordForm.vue
```diff
@@ -127,6 +126,13 @@ export default defineComponent({
       })
     }
 
+    onMounted(() => {
+      setTimeout(() => {
+        const input = passwordInput.value?.$el?.querySelector('input')
+        input?.focus()
+      }, 300)
```
adamantmm, line 133: Why 300? 
Looks like a crutch.

skranee: I'm just testing rn, it's a draft for now, I need to push the changes in order to test it on my phone. In a regular browser on the laptop it works fine. I'm trying to figure out how to avoid iPhone's zoom on the input

skranee:
Okay, seems like putting the font-size strictly to 16px helped. I've tested on safari and google on iPhone 11 and everything works fine

adamantmm:
> Okay, seems like putting the font-size strictly to 16px helped. I've tested on safari and google on iPhone 11 and everything works fine

It was 16px earlier, but you've set it strictly?

skranee:
> It was 16px earlier, but you've set it strictly?
>
Yeah, I've set that manually again with !important and it helps. Found that on stackoverflow, seems like this kinda troubles with iOS are not really new.

src/components/LoginForm.vue
```diff
@@ -165,6 +165,18 @@ export default defineComponent({
 @use '@/assets/styles/settings/_colors.scss';
 @use 'vuetify/settings';
 
+.v-text-field {
```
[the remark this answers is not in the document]
bludnic: Same for `src/components/LoginPasswordForm.vue`

bludnic (approved):

dvelikiy:
[Bug/UI: Markdown doesn't support headers](https://trello.com/c/mVRwZ6ik/743-bug-ui-markdown-doesnt-support-headers)

PR #765, opened 2025-04-16
feat: display getting public key in chatroom

skranee:
It's a draft for now, need to finish little things.


src/views/Chats.vue
```diff
@@ -146,7 +145,11 @@ export default {
     this.destroyScrollListener()
   },
   methods: {
-    openChat(partnerId, messageText) {
+    openChat(partnerId, messageText, retrieveKey = false, partnerName) {
```
S-FrontendDev, line 148: not good to add var with default value before requried

src/components/AChat/AChat.vue
```diff
@@ -63,201 +66,189 @@ const classes = {
...
+    loading: boolean
+    locale: string
+  }>(),
+  {
+    messages: () => [],
+    partners: () => [],
+    loading: false,
+    locale: 'en'
+  }
+)
```
S-FrontendDev, line 83: suggestion:

interface AChatProps {
  messages: NormalizedChatMessageTransaction[]
  partners: User[]
  userId: string
  loading: boolean
  locale: string
}

const props = withDefaults(defineProps<Props>(), {
  messages: () => [],
  partners: () => [],
  loading: false,
  locale: 'en'
})

src/components/AChat/AChat.vue
```diff
@@ -63,201 +66,189 @@ const classes = {
...
+  classes,
+  messagesRef,
+  maintainScrollPosition,
+  scrollToBottom,
+  scrollTo,
+  scrollToMessage,
+  scrollToMessageEasy,
+  getSenderMeta,
+  isScrolledToBottom
+})
```
S-FrontendDev, line 253: ??

what purpose?

src/components/AChat/AChatForm.vue
```diff
@@ -42,259 +43,253 @@
...
-                e.preventDefault()
-                this.submitMessage()
-              }
-            }
+const props = withDefaults(defineProps<Props>(), {
+  partnerId: '',
+  messageText: '',
+  showSendButton: true,
+  sendOnEnter: true,
+  label: 'Type a message',
```
S-FrontendDev, line 69: is it correct to add message for one locale?

src/components/Chat/ChatPlaceholder.vue
```diff
@@ -0,0 +1,119 @@
...
+<script lang="ts" setup>
+import Logo from '@/components/icons/common/Logo.vue'
+import { useI18n } from 'vue-i18n'
+
+const { t } = useI18n()
+
+defineProps<{
+  showPlaceholder: boolean
+  isGettingPublicKey: boolean
+  isKeyMissing: boolean
```
S-FrontendDev, line 33: can you add type? I think we don't need mix props definition

src/components/Chat/Chat.vue
```diff
@@ -83,6 +83,14 @@
         </chat-toolbar>
       </template>
 
+      <template #placeholder>
+        <chat-placeholder
+          :show-placeholder="showNewChatPlaceholder"
```
S-FrontendDev, line 88: may be we need add v-if?

S-FrontendDev: or v-model

src/components/Chat/ChatPlaceholder.vue
```diff
@@ -0,0 +1,119 @@
...
+  <div v-if="showPlaceholder || isGettingPublicKey || isKeyMissing" :class="classes.root">
+    <div v-if="showPlaceholder && !isKeyMissing" :class="classes.container">
+      <Logo style="width: 35px" />
+      <p v-for="key in keys" :key="key" :class="classes.row">
+        {{ t(`chats.placeholder.${key}`) }}
+      </p>
+    </div>
+    <div
+      v-if="isGettingPublicKey || isKeyMissing"
+      :class="[classes.container, `${classes.container}_public-key`]"
```
S-FrontendDev, line 11: what purpose to second class?

src/components/AChat/AChatForm.vue
adamant-al: I think no, we have i18n for this

src/components/ChatStartDialog.vue
```diff
@@ -136,20 +136,13 @@ export default {
...
-        .then((_publicKey) => {
-          this.$emit('start-chat', this.recipientAddress, this.uriMessage)
-          this.show = false
-        })
-        .catch((err) => {
-          this.$store.dispatch('snackbar/show', {
-            message: err.message // @todo translations
-          })
-        })
+      return this.$emit(
```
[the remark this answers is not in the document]
skranee: Technically not, thought it'd be better for readability, but can remove that

src/components/AChat/AChat.vue
skranee: Some of these methods are used in other components. While using composition API, they're not visible in another component through refs (if I'm not mistaken), so in order to fix that, you need to expose methods and vars you want to use through refs.

src/components/AChat/AChatForm.vue
skranee: It's how it was, I only converted that to composition API. I'll make a locale var for that though

src/components/Chat/ChatPlaceholder.vue
skranee: like in the previous suggestion?

src/components/Chat/Chat.vue
skranee: ChatPlaceholder component contains a couple of components that have logical relation. This prop does not decide the visibility of the entire component, but only one part of it. So we can not use v-if

src/components/Chat/ChatPlaceholder.vue
skranee: Basic styles of first css class + styles of modificator of that class

src/components/AChat/AChat.vue
bludnic: What about this approach? https://blog.vuejs.org/posts/vue-3-5

It looks cleaner, and I think we can even keep the type inside the generic. Defining it outside will make sense for cases when we need to export the Props type outside the component, but that’s rare in our codebase.

image

bludnic: Most cases:

```ts
const { count = 0, msg = 'hello' } = defineProps<{
  count?: number
  message?: string
}>()
```

Special cases:

```ts
export type Props {
  count?: number
  message?: string
}

const { count = 0, msg = 'hello' } = defineProps<Props>()
```

src/components/AChat/AChat.vue
S-FrontendDev: check full object, i think you need couple methods, not all

src/components/AChat/AChat.vue
bludnic: More about props destructure in Vue 3.5+ https://vuejs.org/guide/components/props.html#reactive-props-destructure

src/components/Chat/ChatPlaceholder.vue
S-FrontendDev: but modificator without any conditions

src/components/AChat/AChat.vue
S-FrontendDev: i think we need to use last(special) variant. Because in future we can move types to separate file or export them from component for tests, additional types and etc

src/components/Chat/Chat.vue
S-FrontendDev: v-model will be better then

src/components/AChat/AChat.vue
bludnic: Don't see a reason the Props type to be moved to a separate file (it belongs to the component). Exporting for tests make sense though.

skranee: > Most cases:
> 
> ```ts
> const { count = 0, msg = 'hello' } = defineProps<{
>   count?: number
>   message?: string
> }>()
> ```
>
I don't really think it's more readable than ```withDefaults``` variant tbh. ```withDefaults``` instantly says you by name that you have default settings for some of the props, it's easier to read and to find in the project imo.

skranee: So I vote either for typing Props or setting them in the ```defineProps``` right away. But since you've previously discussed that and chose to use typing, I suggest using types in every place, even for little props in order to maintain common codestyle. I don't think it's a good idea to export Props though. Yeah, it's good for tests, but we'll have tens of ```export type Props``` in a single project, is it good?

src/components/AChat/AChat.vue
skranee: Yeah, some of them are not needed, you're right

src/components/Chat/Chat.vue
skranee: Why? Isn't current variant better to read and quickly understand what var is used for?

src/components/Chat/ChatPlaceholder.vue
skranee: Yes, because condition is in ```v-if```. It's the same container as is used above in the file, but with little changes, you don't always need a straightforward condition for a modificator, do you?

src/components/Chat/Chat.vue
S-FrontendDev: ok i thougth it's main var for showing

src/components/Chat/ChatPlaceholder.vue
S-FrontendDev: @bludnic do we need to add it to classes object?

src/components/ChatStartDialog.vue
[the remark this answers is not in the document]
S-FrontendDev: emit doesn't have returning value, i think it's not readability

NikIvv (commented):
image

I'd prefer not mixing up Composition and Options API, unless the Chats.vue is already planned to be changed into Composition API style in one of the latter commits.

skranee:
> image
> 
> I'd prefer not mixing up Composition and Options API, unless the Chats.vue is already planned to be changed into Composition API style in one of the latter commits.

It's changed to composition here
https://github.com/Adamant-im/adamant-im/pull/746

src/components/Chat/Chat.vue
```diff
@@ -472,9 +472,13 @@ onMounted(async () => {
...
   window.removeEventListener('keyup', onKeyPress)
   Visibility.unbind(Number(visibilityId.value))
+
+  if (isNewChat.value) {
+    store.commit('chat/removeNewChat', props.partnerId)
+  }
 })
 
-async function retrievePublicKey(partnerId: string, partnerName: string) {
+async function createChat(partnerId: string, partnerName: string) {
```
S-FrontendDev, line 505: We need to decide on the style of function declaration. Because in some places these are arrow functions, and in other places they are just functions. As far as I understand, in most new components we have arrow functions

skranee: I try to use ```function``` declaration when I write some method that can be used in a component and arrow functions for some little functions like setters and getters and obviously callbacks

skranee:
Chats.vue is still in options API, we need to merge this one first https://github.com/Adamant-im/adamant-im/pull/746 (here Chats.vue is in composition and with resolved conflicts), so then I'll quickly convert it to composition.

adamantmm:
@skranee 
#746 is merged

skranee:
[(Mirrored) {Connection} Feat: Starting a new chat placeholder](https://trello.com/c/dI4kf5GV/690-mirrored-connection-feat-starting-a-new-chat-placeholder)

skranee:
[(Mirrored) {Connection} Feat: Starting a new chat placeholder](https://trello.com/c/dI4kf5GV/690-mirrored-connection-feat-starting-a-new-chat-placeholder)

skranee:
It seems like the only thing left to do is to fix this weird bug with scrollbar, for now I have absolutely no idea why this happens and how to fix this little thing even though I've spent hours on trying to discover that. If you have any suggestions - I'd be highly greatful for them because I consider that bug annoying and worth of being fixed unless Alexey (@adamant-al) tells the opposite and desides that this bug is not a priority right now.

adamantmm:
@skranee 
Is it still a draft? If not, change status and ask to review.

skranee:
> @skranee Is it still a draft? If not, change status and ask to review.

A draft for now

src/locales/en.json
```diff
@@ -33,6 +33,18 @@
     "ok": "OK",
     "partner_info": "Partner info",
     "partner_name": "Partner name",
+    "placeholder": {
+      "anonymous": "\uD83D\uDD76\uFE0F Anonymous: partners never see your IP address",
```
dvelikiy, line 37: image
image

spy emoji is more visible

src/components/Chat/Chat.vue
```diff
@@ -444,11 +470,35 @@ watch(replyMessageId, (messageId) => {
   })
 })
 
+watch(userMessages, () => {
+  // isFirstCallSkipped needed in order to properly handle reloading of a chat where the last
+  // message was from the partner and not you (do not show placeholder)
+  if (isFulfilled.value && isFirstCallSkipped.value) {
+    showNewChatPlaceholder.value = !userMessages.value.length
```
dvelikiy, line 477: `userMessages.value.length === 0` is a bit more readable

dvelikiy (approved):

src/locales/en.json
skranee: Depends on what theme you have, I like glasses more on light theme for example. I think it's more a matter of taste

src/components/Chat/Chat.vue
skranee: Can't really agree on that one, seems like a matter of taste again, but I might be wrong

adamant-al:
@S-FrontendDev @bludnic 
Ps review

src/components/AChat/AChat.vue
```diff
@@ -60,204 +63,222 @@ const classes = {
...
-    const emitScroll = throttle(
-      () => emit('scroll', currentScrollTop.value, isScrolledToBottom()),
-      200
-    )
+const emit = defineEmits(['scroll', 'scroll:bottom', 'scroll:top'])
 
-    const attachScrollListener = () => {
-      messagesRef.value?.addEventListener('scroll', onScroll)
-    }
+const messagesRef = ref<HTMLDivElement | null>(null)
```
S-FrontendDev, line 87: tempalateRef?

src/components/AChat/AChatForm.vue
```diff
@@ -42,268 +43,274 @@
...
+const isInputFocused = ref(false)
+const messageTextarea = ref()
+
+const className = 'a-chat'
+const classes = [
+  `${className}__form`,
+  {
+    [`${className}__form--is-active`]: !!message.value
+  }
+]
```
S-FrontendDev, line 100: it's not right way to use classes

src/components/AChat/AChat.vue
```diff
@@ -10,13 +10,16 @@
           <v-progress-circular
             v-show="loading"
             indeterminate
-            color="primary"
-            size="24"
+            :class="[classes.spinner, !messages.length ? `${classes.spinner}_no-messages` : '']"
```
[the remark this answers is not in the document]
skranee: Just another way of setting class names, I think it's more a matter of taste

src/components/AChat/AChat.vue
```diff
@@ -60,204 +63,222 @@ const classes = {
...
+})
+
+defineExpose({
+  scrollToBottom,
+  isScrolledToBottom,
+  scrollToMessageEasy,
+  maintainScrollPosition,
+  scrollToMessage,
+  scrollTo
+})
```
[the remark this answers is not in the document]
skranee: yes

src/components/AChat/AChatForm.vue
```diff
@@ -42,268 +43,274 @@
...
+
+const isEmojiPickerOpen = computed({
+  get: () => chatStateStore.isEmojiPickerOpen,
+  set: setEmojiPickerOpen
+})
+const placeholder = computed(() => props.label ?? t('chats.type_a_message'))
+const isDesktopDevice = computed(() => !isMobile())
```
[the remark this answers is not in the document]
skranee: why not?

src/components/AChat/AChatForm.vue
skranee: why not? they're used as an array, seems fine to me

```<div :class="classes">```

src/components/AChat/AChatForm.vue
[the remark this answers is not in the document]
S-FrontendDev: it's just primitive without reactivity. You don't need create computed for it, just 
`const isDesktopDevice = !isMobile()`

src/components/AChat/AChatForm.vue
S-FrontendDev: in app we reserved classes for reusable list. Let's rename for classList for example

src/components/AChat/AChat.vue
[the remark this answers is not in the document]
S-FrontendDev: for me class object more readable, it's vue best practice

S-FrontendDev (approved):

PR #767, opened 2025-04-20
feat: text does not reach icon and is centered

skranee:
Lmk what you think of this left margin, it's a simplier solution and also makes text to seem more centered than the case when it's longer on the left side than on the right imo.
image


src/components/LoginForm.vue
```diff
@@ -174,7 +174,8 @@ export default defineComponent({
 
     &:deep(.v-field__input) {
       width: 100%;
-      padding-right: 0;
+      margin-right: 28px;
```
S-FrontendDev, line 177: I think padding is preferable for internal

skranee:
[Bug/UI: Login by password (not passphrase) field don't fit 100% width iOS device + Center "Password" label](https://trello.com/c/jMMZOxBB/757-bug-ui-login-by-password-not-passphrase-field-dont-fit-100-width-ios-device-center-password-label)

bludnic (approved):

adamant-al (approved):

PR #774, opened 2025-04-24
fix: formatting for markdown headings in chat list and notifications

dvelikiy:
Please check using this text:

```
# Header 1
Simple # text

# should remain only for 'Simple # text'
```

### DIFF
--- a/src/lib/markdown.ts
+++ b/src/lib/markdown.ts
@@ -67,23 +67,30 @@ export function renderMarkdown(text = '') {
  * @param {string} text text to process
  * @returns {string} resulting clear text of the first line
  */
-export function removeFormats(text = '') {
-  const node = document.createElement('div')
-  const textWithSymbol = text.replace(/\n/g, '↵ ')
-  node.innerHTML = marked.parse(sanitizeHTML(textWithSymbol), { async: false })
-
-  return node.textContent || node.innerText || ''
-}
-
-export function formatMessage(text = '') {
+export function formatMessageBasic(text = '') {
   const node = document.createElement('div')
 
   const textWithSymbol = text.replace(/\n/g, LINE_BREAK_VISUAL)
   node.innerHTML = marked.parse(sanitizeHTML(textWithSymbol), { async: false })
 
-  const textWithoutHtml = node.textContent || node.innerText || ''
+  let textWithoutHtml = node.textContent || node.innerText || ''
+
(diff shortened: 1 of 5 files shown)


skranee (approved):

src/lib/markdown.ts
```diff
@@ -67,23 +67,24 @@ export function renderMarkdown(text = '') {
...
     .replace(LINE_SEPARATOR, LINE_BREAK_VISUAL)
-    .replace(/↵/g, '<span class="arrow-return">↵</span>')
+    .split(LINE_BREAK_VISUAL)
+    // strip all markdown heading signs (#) but only if they are at the beginning of the line 
+    .map((line) => line.replace(/^[#]+ /, ''))
+    .join(LINE_BREAK_VISUAL)
+
+  if (!isNotification) {
+    textWithoutHtml = textWithoutHtml.replace(/↵/g, '<span class="arrow-return">↵</span>')
+  }
```
bludnic, line 87: Maybe we should keep `removeFormats` to avoid introducing this additional check? @dvelikiy

bludnic: @skranee @S-FrontendDev please join the discussion

dvelikiy:
[Bug/UI: Markdown doesn't support headers](https://trello.com/c/mVRwZ6ik/743-bug-ui-markdown-doesnt-support-headers)

src/lib/markdown.ts
dvelikiy: I felt that this is only a minor performance decrease so it could be kept as is. But I could also create a wrapper function removeFormats that will encapsulate this additional logic and be used separately.

skranee: I don't think additinal check is going to be a big problem here, but may be we should swap code in and outside the `if` (and switch the condition to `isNotification` instead of `!isNotification`) block to decrease the usage of `replace.split.map.join`, will it not be a decent improvement?

dvelikiy: @skranee I don't see how we can decrease the usage of `replace.split.map.join` as it's used for both chat previews and notifications. So I've kept it as is.

@bludnic I've created two different functions to separate those concerns.

Please let me know what you guys think.

skranee (approved):

adamantmm (approved):
image

image

S-FrontendDev (approved):

PR #778, opened 2025-04-26
Bug: Enable Hash router for self-hosted (Static Website Hosting)

graycraft:


graycraft:
[Bug: Enable Hash router for self-hosted (Static Website Hosting)](https://trello.com/c/TpH7QH3Q/737-bug-enable-hash-router-for-self-hosted-static-website-hosting)

.github/workflows/github-pages.yml
```diff
@@ -35,6 +38,7 @@ jobs:
 
       - name: Build PWA 🛠️
         run: |
+          echo 'VITE_ROUTER_HISTORY_MODE="hash"' >> .env
```
bludnic, line 41: Move `VITE_ROUTER_HISTORY_MODE` below, under `env:` property. It should work without this hack.

bludnic: Also, you can remove `env` declaration on line 15.

https://github.com/Adamant-im/adamant-im/pull/778/files#diff-983e80aff18936dd64301171c8eec0d019a56ed0e2849985c5ba1c7cb48a8607R15-R17

src/router/index.js
```diff
@@ -183,7 +183,11 @@ const routes = [
 
 const router = createRouter({
   history:
-    process.env.VUE_APP_ELECTRON_MODE === 'production'
+    /**
+     * For GitHub Pages deployment only.
+     * If variable is not available in process environment, use from the `.env` file instead.
+     */
+    (window.__VITE_ROUTER_HISTORY_MODE__ ?? import.meta.env.VITE_ROUTER_HISTORY_MODE === 'hash')
```
bludnic, line 190: Looks like overengineering. What's the purpose of `window.__VITE_ROUTER_HISTORY_MODE__`? Why not just using `import.meta.env.VITE_ROUTER_HISTORY_MODE`?

vite-pwa.config.ts
```diff
@@ -1,12 +1,16 @@
...
-export default mergeConfig(
-  viteBaseConfig,
-  defineConfig({
+export default defineConfig(() => {
+  /** @see https://vite.dev/guide/api-javascript.html#mergeconfig */
+  return mergeConfig(viteBaseConfig, {
+    /** @see https://vite.dev/config/shared-options.html#define */
+    define: {
+      __VITE_ROUTER_HISTORY_MODE__: JSON.stringify(process.env.VITE_ROUTER_HISTORY_MODE)
+    },
```
bludnic, line 13: All the changes in that file will affect only the PWA. We have also Android and Electron configurations that extends the base config https://github.com/Adamant-im/adamant-im/blob/dev/vite-base.config.ts

1. Move the changes to `vite-base.config.ts`
2. Use same approach as for `VITE_PUBLIC_PATH`. I don't think we should to use `define` here.

bludnic (changes requested):

.github/workflows/github-pages.yml
graycraft: https://github.com/Adamant-im/adamant-im/pull/778/commits/054360bfd0f6c162a8b18f0ffac1542fd9b3a35f

src/router/index.js
graycraft: `window.__VITE_ROUTER_HISTORY_MODE__` is from environment variable
`import.meta.env.VITE_ROUTER_HISTORY_MODE` is from the `.env` file
If `.env` file is not an option, it is not necessary

vite-pwa.config.ts
graycraft: Do we really need hash routing for Android and Electron?

src/router/index.js
bludnic: Can you just use `import.meta.env.VITE_ROUTER_HISTORY_MODE` and remove `window.__VITE_ROUTER_HISTORY_MODE__` everywhere?

vite-pwa.config.ts
bludnic: For Electron yes. For other builds, excluding this one, no. We can control the behaviour with `VITE_ROUTER_HISTORY_MODE` anyway.

src/router/index.js
graycraft: Currently no because `VITE_ROUTER_HISTORY_MODE` is not in the `.env` file, it is [set dynamically](https://github.com/Adamant-im/adamant-im/blob/feature/github-pages-hash-routes/.github/workflows/github-pages.yml#L42) in workflow.
If return `echo 'VITE_ROUTER_HISTORY_MODE="hash"' >> .env` [back](https://github.com/Adamant-im/adamant-im/commit/054360b), `import.meta.env.VITE_ROUTER_HISTORY_MODE` will be sufficient

graycraft: https://github.com/Adamant-im/adamant-im/pull/778/commits/af83217b27d4849b08a8fa805cea564294da09a3

graycraft: Strange, it did not work in my profile deployments, so much the better

vite-pwa.config.ts
graycraft: As we resolved [issue](https://github.com/Adamant-im/adamant-im/pull/778#discussion_r2061472958), perhaps this workaround only needed for Electron builds

bludnic (approved):

PR #779, opened 2025-04-27
fix: electron linux build get path by bash

S-FrontendDev:
getting path by js is not working in github. Rollback to bash


S-FrontendDev:
[Feat/Code: Build Linux app so that it run without tweaks](https://trello.com/c/NJlisXN5/746-feat-code-build-linux-app-so-that-it-run-without-tweaks)

skranee (approved):

PR #781, opened 2025-04-30
fix: potential fix favicon error in console on chat page

Linhead:
Mark fix as potential as I didn't manage to recreate the bug on localhost. So I can't check whether it really fixes the problem but it looks like it might help.


Linhead:
[Bug: No favicon error on adm.im/msg](https://trello.com/c/nfSO42n6/716-bug-no-favicon-error-on-admim-msg)

PR #801, opened 2025-06-14
fix: open correct nodes tab

skranee:


skranee:
[{Connection} Feat: Handle wallets with no connection/nodes](https://trello.com/c/LYgyoWJJ/771-connection-feat-handle-wallets-with-no-connection-nodes)

src/views/Home.vue
```diff
@@ -137,6 +141,10 @@ const wallets = computed(() => {
   })
 })
 
+const areNodesDisabled = (crypto: CryptoSymbol) => {
+  return crypto === 'ADM' ? allAdmNodesDisabled.value : allCoinNodesDisabled.value
+}
+
 const updateBalances = () => {
   if (allCoinNodesDisabled.value) {
```
Linhead, line 150: Maybe it will be more straightforward to use wallet-specific node checking with areNodesDisabled? Something like this: 
if (areNodesDisabled(currentWallet.value)) {
...

src/components/nodes/NodesTable.vue
```diff
@@ -150,11 +150,13 @@ const preferFasterServiceNodeOption = computed<boolean>({
   }
 })
 
-onMounted(() => {
-  if (history.state.tab) {
-    tab.value = history.state.tab
-  }
-})
+watch(
+  () => store.state.options.currentNodesTab,
```
S-FrontendDev: what do you think about adding getter for currentNodesTab?

src/views/Home.vue
```diff
@@ -162,21 +170,28 @@ const goToTransactions = (crypto: string) => {
   })
 }
 
-const goToCoinNodes = () => {
+const goToCoinNodes = (tab: string) => {
```
S-FrontendDev: Tab type?

src/views/Home.vue
```diff
@@ -162,21 +170,28 @@ const goToTransactions = (crypto: string) => {
...
 }
 
-const handleBalanceClick = (crypto?: string) => {
-  if (crypto && !allCoinNodesDisabled.value) {
-    return goToTransactions(crypto)
+const handleBalanceClick = (crypto: string) => {
+  const isAdm = crypto === 'ADM'
+
+  if (isAdm && allAdmNodesDisabled.value) {
+    return goToCoinNodes('adm')
```
S-FrontendDev: lest add move to `as const` object and get from it

src/views/Home.vue
```diff
@@ -162,21 +170,28 @@ const goToTransactions = (crypto: string) => {
...
   })
 }
 
-const handleBalanceClick = (crypto?: string) => {
-  if (crypto && !allCoinNodesDisabled.value) {
-    return goToTransactions(crypto)
+const handleBalanceClick = (crypto: string) => {
+  const isAdm = crypto === 'ADM'
+
+  if (isAdm && allAdmNodesDisabled.value) {
```
S-FrontendDev: is it more readable? If not then resolve it

```
const targetType = isAdm ? 'adm' : 'coins';
const isDisabled = isAdm ? allAdmNodesDisabled.value : allCoinNodesDisabled.value;

if (isDisabled) {
  return goToCoinNodes(targetType);
}
```

src/components/nodes/types.ts
```diff
@@ -0,0 +1,8 @@
+export type Tab = 'adm' | 'coins' | 'services' | 'ipfs'
+
+export const TABS = {
+  adm: 'adm',
+  coins: 'coins',
+  services: 'services',
+  ipfs: 'ipfs'
+} as const
```
S-FrontendDev:
```
export const TABS = {
  adm: 'adm',
  coins: 'coins',
  services: 'services',
  ipfs: 'ipfs'
} as const

export type Tab = keyof typeof TABS
```

S-FrontendDev: sorry, for values

export type Tab = typeof TABS[keyof typeof TABS]

Linhead (approved):

src/views/Home.vue
skranee: I tried that. If I'm not mistaken the problem was that I could not discover correct specifications of coins and their nodes. May be there's some but I didn't find it

S-FrontendDev (approved):

PR #803, opened 2025-06-17
fix: android options spinner

skranee:


skranee:
[{Connection} Feat: Show loading spinner when restore from background — PWA](https://trello.com/c/xkUCTzhD/590-connection-feat-show-loading-spinner-when-restore-from-background-pwa)

android/app/src/main/AndroidManifest.xml
```diff
@@ -1,36 +1,37 @@
...
-            <meta-data android:name="android.support.FILE_PROVIDER_PATHS" android:resource="@xml/file_paths" />
-        </provider>
-    </application>
-
-    <!-- Permissions -->
-
-    <uses-permission android:name="android.permission.INTERNET" />
-    <uses-permission android:name="android.permission.CAMERA" />
-</manifest>
+<?xml version="1.0" encoding="utf-8" ?>
```
adamantmm, line 1: Why do we have so many changes? But actually changed few lines.

skranee: Idk tbh, I only added one line. May be the line ending changed so it was registered as a change

skranee:
```
<uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
```
This is the permission I added so the connection status is discoverable

PR #804, opened 2025-06-23
fix: chat-toolbar styles

S-FrontendDev:


S-FrontendDev:
[Bug/UI: Back button left padding is wrong](https://trello.com/c/9fskRocE/790-bug-ui-back-button-left-padding-is-wrong)

PR #809, opened 2025-06-30
fix: correct token data merging from all 4 sources in adamant-wallets…

Linhead:
Update parsing scripts, so that app merges all 4 sources for tokens


Linhead:
https://trello.com/c/CAdzOu7N/795-refactor-code-update-the-script-for-adamant-wallets-copied-from-ios

PR #814, opened 2025-07-07
Fix: remove question sign logic and don't show ... while loading balances

skranee:


skranee:
[{Connection} Feat: Handle wallets with no connection/nodes](https://trello.com/c/LYgyoWJJ/771-connection-feat-handle-wallets-with-no-connection-nodes)

PR #825, opened 2025-07-23
Fix github pages

S-FrontendDev:


graycraft:
Still capturing 404 with two files:
Screenshot from 2025-07-23 18-10-30
Screenshot from 2025-07-23 18-11-11
Screenshot from 2025-07-23 18-11-33

graycraft:
Cleared cache for opened link from GitHub Actions and seems like 404 resolved now.
Anyway, I would strongly recommend to use much more simple approaches, namely with the `<base>` tag.
Instead of determining URL for each resource by JS and inserting `joinUrl(import.meta.env.BASE_URL, '/path/to/resource')` manually in every place

graycraft:
Experiencing an issue with scroll:
Screenshot from 2025-07-23 18-33-39
Screenshot from 2025-07-23 18-34-32

vite-config/manifest.ts
```diff
@@ -7,53 +12,53 @@ export const manifest: Partial<ManifestOptions> = {
     'ADAMANT is a decentralized anonymous messenger based on the blockchain system. It’s independent of any governments or corporations, and even developers due to the distributed network infrastructure that contains an open-source code.',
   icons: [
     {
-      src: '/img/icons/android-chrome-36x36.png',
+      src: joinUrl(BASE, 'img/icons/android-chrome-36x36.png'),
```
bludnic, line 15: Why not using `import.meta.env.BASE_URL`?

bludnic (changes requested):

vite-config/manifest.ts
graycraft: He is caching it in the variable and I think `import.meta.env.BASE_URL` is unavailable in the config

skranee (approved):

vite-config/manifest.ts
bludnic: Is that the case? @S-FrontendDev

bludnic: If yes, just resolve the conversation

adamantmm (approved):

vite-config/manifest.ts
S-FrontendDev: VITE_PUBLIC_PATH was undefined

PR #831, opened 2025-07-27
fix: attachments not downloading except for the first one

skranee:
When sending a message with multiple files, they couldn't be downloaded except for the first one.
Now it's fixed and files also are uplaoded consistently (and files that were uploaded successfully do not reupload after connection is back)

### DIFF
--- a/src/components/AChat/AChatAttachment/AChatFile.vue
+++ b/src/components/AChat/AChatAttachment/AChatFile.vue
@@ -80,7 +80,6 @@ import IconFile from '@/components/icons/common/IconFile.vue'
 import { useStore } from 'vuex'
 import { AChatFileLoader } from './AChatFileLoader'
 import { mdiImageOff } from '@mdi/js'
-import { useConsiderOffline } from '@/hooks/useConsiderOffline'
 
 const className = 'a-chat-file'
 const classes = {
@@ -121,9 +120,6 @@ const isImage = computed(() => {
   return ['jpg', 'jpeg', 'png'].includes(props.file.extension!)
 })
 
-const { consideredOffline } = useConsiderOffline()
-
-
 const fileName = computed(() =>
   isLocalFile(props.file) ? props.file.file.name : props.file.name || 'UNNAMED'
 )
@@ -150,10 +146,6 @@ const fileSize = computed(() => {
 
 const uploadProgress = computed(() => {
   if (isLocalFile(props.file)) {
-    if (consideredOffline.value) {
-      return 0;
-    }
-
     return store.getters['attachment/getUploadProgress'](props.file.file.cid)
   }
 
--- a/src/store/modules/chat/index.js
+++ b/src/store/modules/chat/index.js
@@ -23,7 +23,7 @@ import {
 } from '@/lib/constants'
 import { isStringEqualCI } from '@/lib/textHelpers'
 import { replyMessageAsset, attachmentAsset } from '@/lib/adamant-api/asset'
-import { uploadFiles } from '../../../lib/files'
+import { uploadFile } from '../../../lib/files'
 import { generateAdamantChats } from './utils/generateAdamantChats'
 import {
   AllNodesDisabledError,
@@ -563,12 +563,13 @@ const mutations = {
     state.noActiveNodesDialog = value
   },
 
-  addPendingMessage(state, { messageId, recipientId, timeout, type, files }) {
+  addPendingMessage(state, { messageId, recipientId, timeout, type, files, cids }) {
     state.pendingMessages[messageId] = {
       recipientId,
       timeout,
       type,
-      files
+      files,
+      cids
     }
   },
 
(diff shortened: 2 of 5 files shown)


src/store/modules/chat/index.js
```diff
@@ -936,37 +937,41 @@ const actions = {
...
+    cids.forEach(([cid]) => {
+      commit('attachment/setUploadProgress', { cid, progress: 0 }, { root: true })
+    })
+
+    const uploadData = await dispatch('uploadConsistently', {
+      files,
+      cids
+    })
+
+    console.debug('Files uploaded', uploadData)
```
S-FrontendDev: do we need it?

src/components/AChat/AChatAttachment/AChatFile.vue
```diff
@@ -150,10 +146,6 @@ const fileSize = computed(() => {
 
 const uploadProgress = computed(() => {
   if (isLocalFile(props.file)) {
-    if (consideredOffline.value) {
-      return 0;
-    }
```
S-FrontendDev, line 155: how we processing this case now? I see in start upload handler. What about lose connection after click 'send' button?

src/store/modules/chat/index.js
```diff
@@ -980,13 +985,11 @@ const actions = {
...
 
-        throw err
+        throw uploadData.error
       }
     }
 
-    for (const [cid] of cids) {
-      commit('attachment/resetUploadProgress', { cid }, { root: true })
-    }
+    console.debug('Updated CIDs after upload', newAsset)
```
S-FrontendDev: ?

src/store/modules/chat/index.js
```diff
@@ -1211,6 +1206,52 @@ const actions = {
...
+        }
+
+        return {
+          newCids: oldCidsMutated,
+          error
+        }
+      }
+    }
+
+    return { newCids: uploadedCids }
```
S-FrontendDev: It it possible to use one foreach? I think a lot of quantity cycles

src/store/modules/chat/index.js
skranee: I don't think so, it just was from the very beginning so I never removed it, I'd like to remove all the logs too

src/components/AChat/AChatAttachment/AChatFile.vue
skranee: Now we keep every needed progress in store and connection plays no role in this.
If we lose connection while uploading:
- Messages that were uplaoded before losing connection won't have progress because they're uplaoded (and they won't be reuploaded after reconnection)
- Messages that were **not** uploaded before losing connection have progress = 0 as it should be

src/store/modules/chat/index.js
skranee: This log was also from the very beginning (I didn't remove any logs that were made by others)

src/store/modules/chat/index.js
skranee: We have only 5 files max so it's not a big deal but I can try to optimize the method

src/store/modules/chat/index.js
```diff
@@ -1211,6 +1206,52 @@ const actions = {
...
+          commit('attachment/setUploadProgress', { cid, progress }, { root: true })
+        })
+
+        commit('attachment/resetUploadProgress', { cid }, { root: true })
+
+        uploadedCids.push(uploadedFile.cids)
+      } catch (error) {
+        commit('attachment/setUploadProgress', { cid, progress: 0 }, { root: true })
+
+        const oldCidsMutated = cids
```
S-FrontendDev: ))))

src/store/modules/chat/index.js
```diff
@@ -1211,6 +1206,52 @@ const actions = {
...
+
+        for (const [index, cid] of uploadedCids.entries()) {
+          oldCidsMutated[index] = cid
+        }
+
+        return {
+          newCids: oldCidsMutated,
```
S-FrontendDev: can we return uploadedCids?

src/store/modules/chat/index.js
```diff
@@ -1128,38 +1133,33 @@ const actions = {
...
+      uploadData = await dispatch('uploadConsistently', {
+        files,
+        cids
       })
-    } catch (err) {
-      for (const [cid] of cidsOld) {
-        commit('attachment/resetUploadProgress', { cid }, { root: true })
-      }
 
+      if (uploadData.error) {
```
S-FrontendDev, line 1159: can we catch error from uploadConsistently method? I don't like uploadData.error

src/store/modules/chat/index.js
```diff
@@ -934,39 +941,40 @@ const actions = {
...
-      commit('updateMessage', {
-        id: messageObject.id,
-        partnerId: recipientId,
-        asset: newAsset
-      })
-      console.debug('Updated CIDs after upload', newAsset)
-    } catch (err) {
-      if (!isAllNodesOfflineError(err)) {
-        if (isAllNodesDisabledError(err)) {
+    if (uploadData.error) {
```
S-FrontendDev, line 975: catch from uploadConsistently

src/store/modules/chat/index.js
```diff
@@ -1211,6 +1212,65 @@ const actions = {
...
+        }
+      } finally {
+        unsubscribe()
+      }
+    }
+
+    return { newCids: uploaded }
+  },
+
+  setUploadProgress({ commit }, { cid, progress }) {
```
S-FrontendDev: can you create arrow function like getProgress?

PR #832, opened 2025-07-28
fix: fix black overflow when light theme is turned on

Linhead:
Fix black color overflow when light theme is turned on
image


Linhead:
https://trello.com/c/KFfOOYil/810-bug-samsung-buttons-cover-app

S-FrontendDev (approved):

PR #839, opened 2025-08-06
fix: change released apk name

S-FrontendDev:


scripts/capacitor/build-android.mjs
```diff
@@ -22,4 +24,11 @@ async function run() {
     '--signing-type apksigner'
   ]
   await $$`cap build android ${buildArgs}`
+
+
+  const version = JSON.parse(readFileSync('package.json', 'utf-8')).version
+  const src = 'android/app/build/outputs/apk/release/app-release-signed.apk'
+  const dest = `android/app/build/outputs/apk/release/ADAMANT.Messenger-${version}-signed.apk`
```
bludnic: It will work only when building APK on a local machine. GitHub Workflow builds the app differently https://github.com/Adamant-im/adamant-im/blob/dev/.github/workflows/capacitor-android.yml#L68-L76

bludnic: I would rather modify the GH Workflow and leave this file unchanged. Note that these changes may break the `npm run android:run` functionality, as the original .APK name was changed.

bludnic: I assume these changes are needed to ensure that the automatically uploaded artifacts in the release are named appropriately.

bludnic (approved):

adamantmm (approved):

PR #844, opened 2025-08-11
chore: update min sdk version

S-FrontendDev:


graycraft:
If we comply with Semver specification:

>  patch version MUST be incremented if only backward compatible bug fixes are introduced. A bug fix is defined as an internal change that **fixes incorrect behavior**.

In this case, instead of bumping the patch, a label must be appended with a build date, build hash, etc: 

+ `4.10.6+20250811`

+ `4.10.6+64d7b871`

<https://semver.org/#spec-item-6>

android/variables.gradle
```diff
@@ -1,5 +1,5 @@
 ext {
-    minSdkVersion = 23
+    minSdkVersion = 35
```
adamantmm, line 2: Is it correct?
We allow old devices, too.

adamantmm:
> If we comply with Semver specification:
> 
> > Patch version MUST be incremented if only backward compatible bug fixes are introduced.
> > A bug fix is defined as an internal change that **fixes incorrect behavior**.
> 
> In this case, instead of bumping the patch, a label must be appended with a build date, build hash, etc:
> 
> * `4.10.6+20250811`
> * `4.10.6+64d7b871`
> 
> https://semver.org/#spec-item-6

Prefer to simplify and use `4.10.7` even if it's not technically correct.

PR #847, opened 2025-08-13
:art: chore: update message input styles

kalpovskii:
Closes: https://trello.com/c/H02HmfAB

* Open “Emoji” and “Plus” on hover automatically (no tap) and close on losing focus
* Make the “Send” button disabled if there is nothing to send
* Add on-hover animation for buttons
* Remove blue line
* Add 4px padding between “Plus” and “Send”

Since we are using the library icons, it’s not possible to control their stroke width. I don’t think switching to custom SVG icons would be the best solution. I’d like to hear your thoughts on this

**UPDATE**:

* The icon turns "primary" (blue) color when hovered over

**UPD2:**

* Ensures the "Send" button is always active, even for new accounts with a zero balance. Clicking the button now triggers the faucet popup modal that explains how to get free coins
* Added blue hover effect to the send icon

**UPD3:**

* Changed icons color from "white" to "darken-1"


src/components/Chat/ChatMenu.vue
```diff
@@ -85,7 +86,7 @@ const router = useRouter()
 const store = useStore()
 const { t } = useI18n()
 const chatStateStore = useChatStateStore()
-
+const { mdAndUp } = useDisplay()
```
kalpovskii: Since there is no hover on mobile devices, this will work only on desktop resolutions; on mobile, it will simply be a tap still

skranee:
> Closes: https://trello.com/c/H02HmfAB
> 
> * Open “Emoji” and “Plus” on hover automatically (no tap) and close on losing focus
> * Make the “Send” button disabled if there is nothing to send
> * Add on-hover animation for buttons
> * Remove blue line
> * Add 4px padding between “Plus” and “Send”
> 
> Since we are using the library icons, it’s not possible to control their stroke width. I don’t think switching to custom SVG icons would be the best solution. I’d like to hear your thoughts on this

may be a wrapper for an icon wil; work. chat gpt suggested me that way, I guess you've already tried that but if not, hope that helps

```
<template>
  <svg
    xmlns="http://www.w3.org/2000/svg"
    viewBox="0 0 24 24"
    :stroke="isDisabled ? 'grey' : 'white'"
    fill="none"
    :stroke-width="2"
    width="28"
    height="28"
  >
    <path :d="mdiSend" />
  </svg>
</template>

<script setup>
import { mdiSend } from '@mdi/js'
</script>
```

src/components/AChat/AChatForm.vue
```diff
@@ -35,11 +35,16 @@
       </template>
       <template v-if="showSendButton" #append-inner>
         <slot name="append" />
-        <v-icon class="a-chat__send-icon" :icon="mdiSend" size="28" />
+        <v-icon
+          class="a-chat__form-send-area"
+          :color="isDisabled ? 'grey' : 'white'"
```
kalpovskii: The icon turns grey when isDisabled === true
image

src/components/AChat/AChatForm.vue
```diff
@@ -371,12 +379,9 @@ defineExpose({
...
 
 .a-chat__form-send-area {
-  position: absolute;
-  bottom: 0;
-  right: 0;
-  width: 50px;
-  height: 50px;
-  cursor: pointer;
+  &:hover {
+    color: map.get(colors.$adm-colors, 'primary') !important;
```
kalpovskii: image

src/components/Chat/ChatEmojis.vue
```diff
@@ -28,8 +29,10 @@ export default {
   },
   emits: ['onChange', 'get-emoji-picture'],
   setup() {
+    const { mdAndUp } = useDisplay()
```
S-FrontendDev: src/hooks/useScreenSize.ts

src/components/AChat/AChatForm.vue
```diff
@@ -35,11 +35,16 @@
       </template>
       <template v-if="showSendButton" #append-inner>
         <slot name="append" />
-        <v-icon class="a-chat__send-icon" :icon="mdiSend" size="28" />
+        <v-icon
+          class="a-chat__form-send-area"
+          :color="sendButtonColor"
```
S-FrontendDev: use class for icon color styling

src/components/AChat/AChatForm.vue
S-FrontendDev: don't need !important if you use classes

src/components/Chat/ChatMenu.vue
S-FrontendDev: check our hook for mobile

src/components/Chat/ChatEmojis.vue
```diff
@@ -53,14 +56,44 @@ export default {
...
+        border-radius: 50%;
+        background: currentColor;
+        opacity: 0;
+        transition: 0.4s;
+        z-index: -1;
+      }
+
+      &:hover::before {
+        opacity: 0.1;
+      }
```
S-FrontendDev: move this rule to the root of style block and don't use theme class in your selector, then you don't need to duplicate it

src/lib/constants/index.ts
```diff
@@ -251,4 +251,12 @@ export const REACT_EMOJIS = {
   PARTY_POPPER: '🎉'
 } as const
 
+/** Validation errors for the chat message */
+export const validationErrors = {
+  emptyMessage: 'EMPTY_MESSAGE',
+  notEnoughFunds: 'NON_ENOUGH_FUNDS',
+  notEnoughFundsNewAccount: 'NON_ENOUGH_FUNDS_NEW_ACCOUNT',
+  messageTooLong: 'MESSAGE_LENGTH_EXCEED'
```
S-FrontendDev: let's use PascalCase

src/components/AChat/AChatForm.vue
S-FrontendDev: use class

S-FrontendDev (approved):

PR #851, opened 2025-08-25
:art: chore: add messages grouping

kalpovskii:
Closes: https://trello.com/c/VEX6lOkS

I've implemented a time-based grouping logic. Messages sent within a 30-second interval now have a reduced bottom margin (from 16px to 4px) to create a more cohesive and compact visual unit

The timestamp is now displayed only on the first message in each group

While I found the visual implementation of 'the bubble solution' less aesthetically pleasing, I chose not to remove it entirely, hence I only added grouping logic and spacing

I'm open to suggestions and alternative approaches regarding "the bubble styling"

image

**UPDATE:**

* Added tail styling exclusively for the last message in a group
* Doubled the border-radius for all messages from 8px to 16px

image


S-FrontendDev:
i agree with @skranee. I think last message with time might be bettee
image


and what about merge attchment and message? Without margin
image


@adamant-al

PR #854, opened 2025-08-29
feat: Feat/add touch id face id passkey

Linhead:
Implemented multi-factor authentication system for ADAMANT Messenger with two new authentication methods alongside existing passphrase and password authentications.

 New Authentication Methods:
 Biometric Authentication (Touch ID/Face ID)
  - Native biometric authentication on mobile devices
  - Uses Capacitor Native Biometric plugin
  - Supports fingerprint and facial recognition
  - Mobile-only feature (unavailable in web browsers)

 Passkey Authentication
  - WebAuthn-based passwordless authentication
  - Uses browser's built-in passkey support
  - Supports hardware security keys and platform authenticators
  - Web-only feature (unavailable in mobile apps)

### DIFF
--- a/android/app/capacitor.build.gradle
+++ b/android/app/capacitor.build.gradle
@@ -11,6 +11,7 @@ apply from: "../capacitor-cordova-android-plugins/cordova.variables.gradle"
 dependencies {
     implementation project(':capacitor-community-file-opener')
     implementation project(':capacitor-filesystem')
+    implementation project(':capgo-capacitor-native-biometric')
 
 }
 
--- a/android/app/capacitor.build.gradle
+++ b/android/app/capacitor.build.gradle
@@ -12,6 +12,7 @@ dependencies {
     implementation project(':capacitor-community-file-opener')
     implementation project(':capacitor-filesystem')
     implementation project(':capgo-capacitor-native-biometric')
+    implementation project(':capacitor-secure-storage-plugin')
 
 }
 
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -35,4 +35,5 @@
     <uses-permission android:name="android.permission.CAMERA" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.VIBRATE" />
+    <uses-permission android:name="android.permission.USE_BIOMETRIC" />
 </manifest>
--- a/android/capacitor.settings.gradle
+++ b/android/capacitor.settings.gradle
@@ -7,3 +7,6 @@ project(':capacitor-community-file-opener').projectDir = new File('../node_modul
 
 include ':capacitor-filesystem'
 project(':capacitor-filesystem').projectDir = new File('../node_modules/@capacitor/filesystem/android')
+
+include ':capgo-capacitor-native-biometric'
+project(':capgo-capacitor-native-biometric').projectDir = new File('../node_modules/@capgo/capacitor-native-biometric/android')
--- a/android/capacitor.settings.gradle
+++ b/android/capacitor.settings.gradle
@@ -10,3 +10,6 @@ project(':capacitor-filesystem').projectDir = new File('../node_modules/@capacit
 
 include ':capgo-capacitor-native-biometric'
 project(':capgo-capacitor-native-biometric').projectDir = new File('../node_modules/@capgo/capacitor-native-biometric/android')
+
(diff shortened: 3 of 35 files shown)


Linhead:
https://trello.com/c/r5W6rcgH/366-feat-add-touch-id-face-id-passkey

adamantmm:
Review note from the security audit work (#958). Not a blocker for the direction — this PR is the right direction, and it is the reason a competing "non-extractable CryptoKey in IndexedDB" design was considered and dropped. Posting now so the point is not lost before the rework that follows v4.12.0 and v4.13.0.

## The passkey is currently a presence check, not a source of key material

`src/lib/auth/PasskeyAuthService.ts` calls `navigator.credentials.create()` and `navigator.credentials.get()` with a random `challenge` and no `extensions`. Nothing is returned from the authenticator that could be used as a secret — the call only establishes that a user was present.

Meanwhile the flow works with an `encryptionKey: Uint8Array` that comes from somewhere else. If that key is stored on the device alongside the ciphertext it protects, then the security property is:

- **the passkey decides whether the app shows the unlock screen**
- **the key needed to decrypt the passphrase is already on disk**

An XSS payload or a malicious extension never sees the unlock screen. It reads the key and the ciphertext directly and decrypts offline. This is structurally the same problem as the one tracked as ADM-SEC-005 in the audit — the secret and the thing that protects it live in the same place.

To be clear about what this does buy: it does raise the bar against someone who picks up an unlocked device and opens the app. That is worth something. It just is not protection against script execution or against profile theft, which is what a passphrase-holding wallet needs.

## What makes it real: the WebAuthn PRF extension

`prf` asks the authenticator to derive a stable secret from its own non-exportable key, released only after user verification. That secret is the thing to derive the encryption key from — it exists nowhere on disk and cannot be recomputed without the authenticator plus the biometric.

Registration:

```js
const credential = await navigator.credentials.create({
  publicKey: {
    challenge,
    rp: { name: 'ADAMANT Messenger' },
    user: { id, name, displayName },
    pubKeyCredParams: [{ type: 'public-key', alg: -7 }],
    authenticatorSelection: { userVerification: 'required', residentKey: 'required' },
    extensions: { prf: {} }
  }
})

// credential.getClientExtensionResults().prf?.enabled  -> is PRF usable at all
```

Unlock:

```js
const assertion = await navigator.credentials.get({
  publicKey: {
    challenge,
    allowCredentials: [{ id: credentialId, type: 'public-key' }],
    userVerification: 'required',
    extensions: { prf: { eval: { first: prfSalt } } }   // prfSalt is stored, not secret
  }
})

const secret = assertion.getClientExtensionResults().prf.results.first  // 32 bytes
// derive the encryption key from `secret`, decrypt the passphrase, zero `secret`
```

`prfSalt` is a per-installation random value and can be stored in the clear — it is a domain separator, not a secret. The authenticator's own key never leaves the Secure Enclave / TPM / hardware key.

## Points worth deciding before the rework

1. **PRF availability is narrower than WebAuthn itself** — roughly Chrome 132+ and Safari 18+. `prf.enabled` has to be checked at registration time, and when it is false the passkey option should either not be offered or be labelled honestly as convenience rather than protection. Silently falling back to a locally stored key is the failure mode to avoid.

2. **The mobile path needs the equivalent treatment.** `BiometricAuthService` via Capacitor Native Biometric has the same question: is the biometric gating a UI transition, or is it releasing a key from Android Keystore / iOS Keychain that was created with a "require user authentication" flag? Only the second protects anything. The plugin's `verifyIdentity()` alone is the first.

3. **Biometry is a session unlock, not a per-action prompt.** One verification at app start yields the key, the passphrase is decrypted into memory, and everything after that works without prompts. Re-prompting is then a product choice — an inactivity timeout, or transfers above a threshold — not a WebAuthn requirement. Worth stating explicitly so the UX discussion does not start from "a fingerprint for every message".

4. **Recovery has to stay.** The passphrase path must remain reachable: a lost device means a lost authenticator, and PRF secrets cannot be exported or backed up.

5. **Storage of the wrapped passphrase.** Whatever holds the ciphertext should carry a versioned descriptor (algorithm, KDF parameters, PRF salt, credential id) so that parameters can be raised later without forcing users to sign in again. The same descriptor pattern is being introduced for the password KDF in v4.12.0.

## Related

- #958 — full security audit; ADM-SEC-005 is the finding this PR ultimately addresses
- ADM-SEC-004 / ADM-SEC-025 in that audit — the password KDF is being reworked in v4.12.0 (per-user salt, scrypt); the passkey path should reuse the same key-derivation and descriptor code rather than growing a second one

PR #861, opened 2025-09-08
:bug: fix: add left type to emojipicker position

kalpovskii:
Closes: https://github.com/Adamant-im/adamant-im/issues/904

* Added a new "left" type for the emojipicker component to ensure the picker is displayed on the top-left side when needed. It prevents it from going outside the screen boundaries

image


S-FrontendDev:
@adamant-al is it ok? Maybe we need to block scroll? 

image

PR #864, opened 2025-09-12
fix: add removing obsolete coins that no longer exist in configs in g…

Linhead:
### Fix: Remove obsolete cryptocurrency configs in wallets generation script

**Summary**

  - Fixed a bug where removed cryptocurrencies remained in config files after running wallets:generate
  - Added logic to clean up obsolete coin entries that no longer exist in the adamant-wallets source

  **Problem**

  The updateConfig() function in scripts/wallets.mjs only added/updated cryptocurrency configurations but never removed obsolete ones. When cryptocurrencies were removed from the adamant-wallets submodule, they would persist indefinitely in the config files (src/config/*.json), leading to:
  - Unnecessary API calls to non-existent nodes
  - Outdated configuration data

  **Solution**

  Enhanced the updateConfig() function to:
  1. Remove obsolete entries: Delete config keys that no longer exist in the source data
  2. Add/update active entries: Maintain existing functionality for current cryptocurrencies

  **Changes**

  - Added cleanup logic before updating configs in scripts/wallets.mjs:179-184
  - Fixed lodash import to use default import (import _ from 'lodash') for consistency


S-FrontendDev:
When I test the script on my branch feat/remove-kly, the command runs successfully.
image

When I test it on your branch, I get an error.

image

Issue #927, opened 2026-02-24
[Task] Remove runtime unsafe-eval sources and resume strict CSP hardening


dev-adamant-im:
## Summary

`unsafe-eval` is still present in runtime bundles, so enforcing strict CSP without `unsafe-eval` is currently unsafe and can break app flows.

This task tracks full removal of runtime `unsafe-eval` sources and safe migration to stricter CSP.

## Related

- Original task: #916
- Context PR branch work: #917

## What Was Done Already

- Added CSP report endpoint `api/csp-report.js` for Vercel report collection
- Ran CSP dry-run in Vercel environments and collected violations
- Investigated bundle sources with sourcemaps for `Function(...)` / `new Function(...)` paths
- Upgraded `vuedraggable` from `2.24.3` to `4.1.0`
- Switched draggable import to source module in `src/views/Wallets.vue` to avoid UMD bundle path

### Confirmed runtime/eval-like sources in bundle

- `core-js/internals/global-this.js` -> `Function('return this')()` fallback
- `lodash-es/_root.js` -> `Function('return this')()` fallback
- `engine.io-client/build/esm/globals.js` -> `Function("return this")()` fallback
- `setimmediate/setImmediate.js` -> `new Function("" + callback)`

### Dependency paths (why they are included)

- `engine.io-client` is included via `socket.io-client` (chat sockets)
- `setimmediate` is pulled by `web3-eth` chain and polyfill stack
- `lodash-es` fallback enters through current imports (`throttle`, `get`, `random`, `cloneDeep`)
- `core-js` fallback is included by current polyfill usage in app bootstrap

### Important conclusion

- Strict CSP `script-src` without `unsafe-eval` should not be enforced yet
- First remove/replace all runtime `unsafe-eval` sources, then re-enable CSP hardening in Report-Only and move to enforce

## Temporary operational decision

To avoid regressions in preview/dev Vercel builds while cleanup is in progress:

- `vercel.json` no longer sets `Content-Security-Policy-Report-Only`
- CSP hardening for Vercel will be re-enabled after source cleanup

### A) Remove runtime `unsafe-eval` sources

- Replace `lodash-es` usage with targeted local utilities where possible
- Evaluate safer socket stack options and `engine.io-client` eval fallback removal path
- Remove `setimmediate` runtime usage from web bundle (web3/polifill path)
- Minimize/replace `core-js` paths that emit eval-style global fallback where feasible

### B) Re-enable CSP hardening safely

[...]

PR #964, opened 2026-08-10
Fix: Complete security hardening part 2

### LINKED ISSUE #928: [Task] Prepare safe ESLint 10 migration without losing import checks
### Summary
Prepare a safe migration path from ESLint 9 to ESLint 10 without reducing current lint coverage for import correctness

### Details
Context and current state
- During dependency upgrade in task #916 we tested ESLint 10 migration (`eslint` + `@eslint/js`)
- The direct upgrade is currently blocked by `eslint-plugin-import` peer range (`^2 || ^3 || ^4 || ^5 || ^6 || ^7.2.0 || ^8 || ^9`)
- `npm install` with ESLint 10 produced peer-resolution conflicts (`ERESOLVE`) in normal mode

What was tested
- Variant A: keep ESLint 10 and remove `eslint-plugin-import` + `eslint-import-resolver-typescript`
- Variant B: rollback to ESLint 9 and keep current import plugin stack
- We used baseline checks in this repo (`npm run lint`, `npm run typecheck`, `npm run test -- --run`)

Findings
- Variant A allows ESLint 10 installation but removes import validation coverage from lint
- Probe `import electron from 'vite-plugin-electron'` was not reported by ESLint after plugin removal
- Probe `const value: string = 123` is not reported by ESLint and is only reported by `vue-tsc`, which is expected for current config
- Variant B restores previous behavior and keeps import checks in pre-commit lint flow

Decision made now
- Keep ESLint 9 stack in repository for stability
- Track ESLint 10 migration as a separate follow-up task

Related
- Parent dependency task: #916

### Checklist
- [ ] Re-check ecosystem support for ESLint 10 in import lint stack (`eslint-plugin-import` or equivalent)
- [ ] Evaluate alternatives and document tradeoffs (`eslint-plugin-import`, `eslint-plugin-import-x`, other maintained options)
- [ ] Prepare migration branch with ESLint 10 and zero `--force` in normal install flow
- [ ] Keep coverage parity for current import checks (`import/default`, `import/named`, unresolved imports with TS resolver)
- [ ] Add regression tests or lint probes that fail if import coverage is accidentally dropped
- [ ] Verify baseline commands pass: `npm run lint`, `npm run typecheck`, `npm run test -- --run`

### Notes
Potential risk if migrated too early
- Silent regressions in import correctness can reach runtime and break build/dev flows
- Example: default import mismatch for `vite-plugin-electron` is not caught without import plugin rules

Potential migration acceptance criteria
- Clean `npm install` without force flags
- Import checks remain enforced in lint and pre-commit
- No new false-positive noise that requires disabling core safety rules globally

### Verification
A candidate ESLint 10 migration is acceptable when all conditions hold
- `npm install` succeeds without `--force` and without peer conflict warnings for lint packages
- `npm run lint` reports import regressions currently caught by ESLint 9 setup
- `npm run typecheck` and `npm run test -- --run` stay green

### LINKED ISSUE #929: [Task] Stale direct dependency review and phased replacement plan
## Summary
Review direct dependencies and devDependencies, identify packages that are stale, and define a safe migration/removal plan without breaking cryptography, protocol compatibility, or build targets

## Scope and methodology
- Checked all direct `dependencies` and `devDependencies` from `/Users/Das/Work/adamant.im/Dev/PWA/adamant-im/package.json`
- Used npm registry metadata to get publish dates for installed versions and latest published versions
- Filtered stale packages by threshold `last release < 2024-02-25` (older than 2 years)
- Verified deprecation and installation noise via `npm i --dry-run --ignore-scripts --no-audit --no-fund`
- Collected security data via `npm audit --json`

## Key findings
- There are **0 direct packages** that are both deprecated and older than 2 years
- There are **43 direct packages** that are older than 2 years and not deprecated
- `npm i` does not produce deprecated warnings for direct packages
- In the stale direct set, only `hdkey` is flagged by `npm audit` (low severity, indirect via `secp256k1`)
- Separate from this stale set, `npm audit` still reports high findings in other direct packages (for example `bitcoinjs-lib`, `ecpair`, `@capacitor/assets`, `@vue/test-utils`)

## Table A: Replace or remove candidates
| Package | Type | Last release | Current usage in repo | Known vulnerabilities | `npm i` noise | Alternative(s) | Recommendation |
|---|---|---:|---|---|---|---|---|
| `simple-audio` | dependency | 2017-06-07 | No imports found | None in audit for this package | None | Remove | Remove now |
| `i` | dependency | 2021-09-21 | No imports found | None in audit for this package | None | Remove | Remove now |
| `jest-serializer-vue` | devDependency | 2022-11-18 | No imports found, test stack is Vitest | None in audit for this package | None | Remove | Remove now |
| `@esbuild-plugins/node-globals-polyfill` | devDependency | 2023-01-27 | No imports found | None in audit for this package | None | `vite-plugin-node-polyfills` (already used) | Remove after build check |
| `@types/core-js` | devDependency | 2023-11-07 | No imports found | None in audit for this package | None | Remove if typecheck passes | Remove after typecheck |
| `@types/lodash-es` | devDependency | 2023-11-21 | No imports found | None in audit for this package | None | Remove if typecheck passes | Remove after typecheck |
| `@types/pbkdf2` | devDependency | 2023-11-07 | No imports found | None in audit for this package | None | Remove if typecheck passes | Remove after typecheck |
| `notifyjs` | dependency | 2017-02-15 | Browser notifications wrapper in `src/lib/notifications.js` | None in audit for this package | None | Native Notifications API wrapper | Migrate |
| `visibilityjs` | dependency | 2018-08-30 | Visibility checks in chat and hooks | None in audit for this package | None | `@vueuse/core` (`useDocumentVisibility`) | Migrate |
| `promise-queue` | dependency | 2018-01-28 | Message and API queues in chat/adamant-api | None in audit for this package | None | `p-queue` or `bottleneck` | Migrate |
| `throttle-promise` | dependency | 2017-11-08 | IndexedDB throttling plugin | None in audit for this package | None | `p-throttle` or local throttle utility | Migrate |
| `vuex-persist` | dependency | 2020-09-27 | Local/session persistence plugins for Vuex | None in audit for this package | None | `pinia-plugin-persistedstate` | Migrate with store plan |
| `vuex` | dependency | 2022-10-14 | Primary legacy store layer | None in audit for this package | None | Pinia | Gradual migration |
| `vuedraggable` | dependency | 2021-08-22 | Wallet list drag-drop in `src/views/Wallets.vue` | None in audit for this package | None | `vue-draggable-plus` or SortableJS wrapper | Migrate |
| `copy-to-clipboard` | dependency | 2022-11-13 | Address/passphrase copy actions | None in audit for this package | None | `navigator.clipboard` + fallback | Migrate |
| `b64-to-blob` | dependency | 2017-08-21 | QR export conversion | None in audit for this package | None | `fetch(dataUrl).blob()` | Migrate |
| `file-saver` | dependency | 2020-11-19 | QR image save flow | None in audit for this package | None | Native `URL.createObjectURL` + `<a download>` | Migrate |
| `detect-browser` | dependency | 2021-12-04 | Mobile/browser detection hooks | None in audit for this package | None | Feature detection + `userAgentData` fallback | Migrate |
| `hdkey` | dependency | 2023-01-18 | ETH key derivation (`src/lib/eth-utils.js`) | Low in audit via `secp256k1` | None | `@scure/bip32` | Migrate with vector tests |

## Table B: Keep for now and monitor
| Package | Type | Last release | Current usage in repo | Known vulnerabilities | `npm i` noise | Alternative(s) | Recommendation |
|---|---|---:|---|---|---|---|---|
| `bytebuffer` | dependency | 2016-02-12 | ADAMANT transaction byte serialization | None in audit for this package | None | No low-risk drop-in | Keep |
| `https-browserify` | dependency | 2017-04-04 | Node polyfill alias in Vite config | None in audit for this package | None | Remove only after polyfill cleanup | Keep |
| `os-browserify` | dependency | 2017-04-20 | Node polyfill alias in Vite config | None in audit for this package | None | Remove only after polyfill cleanup | Keep |
| `sodium-browserify-tweetnacl` | dependency | 2019-04-19 | Core crypto implementation in `adamant.js` | None in audit for this package | None | Migration is high-risk | Keep |
| `rollup-plugin-node-polyfills` | dependency | 2019-06-14 | Polyfill stack in Vite build | None in audit for this package | None | Replace only in dedicated polyfill refactor | Keep |
| `ed2curve` | dependency | 2020-01-16 | Key conversion in crypto modules | None in audit for this package | None | Few compatible alternatives | Keep |
| `tweetnacl` | dependency | 2020-02-10 | `tweetnacl/nacl-fast` usage in crypto code | None in audit for this package | None | Migration requires cryptographic validation | Keep |
| `path-browserify` | dependency | 2020-03-03 | Node polyfill alias in Vite config | None in audit for this package | None | Remove only after polyfill cleanup | Keep |
| `stylefire` | dependency | 2020-03-31 | Chat animation utility | None in audit for this package | None | CSS/Web Animations API | Keep |
| `stream-browserify` | dependency | 2020-04-16 | Node polyfill alias in Vite config | None in audit for this package | None | Remove only after polyfill cleanup | Keep |
| `buffer` | dependency | 2020-11-23 | Required by browser crypto and attachment code | None in audit for this package | None | Remove only after polyfill cleanup | Keep |
| `stream-http` | dependency | 2021-04-14 | Node polyfill alias in Vite config | None in audit for this package | None | Remove only after polyfill cleanup | Keep |
| `coininfo` | dependency | 2022-03-08 | Bitcoin-family network params | None in audit for this package | None | Local network constants | Keep |
| `lorem-ipsum` | devDependency | 2022-06-10 | Test fixtures only | None in audit for this package | None | Static test fixtures | Keep |
| `popmotion` | dependency | 2022-08-15 | Chat animation utility | None in audit for this package | None | CSS/Web Animations API | Keep |
| `bip39` | dependency | 2023-02-25 | Mnemonic and passphrase flows | None in audit for this package | None | `@scure/bip39` | Keep now, review with `hdkey` migration |
| `deepmerge` | dependency | 2023-03-16 | State merge in IDB state layer | None in audit for this package | None | Local merge helper | Keep |
| `mitt` | dependency | 2023-07-04 | Lightweight event bus | None in audit for this package | None | Local emitter | Keep |
| `assert` | dependency | 2023-09-07 | Assertions and polyfill compatibility | None in audit for this package | None | Native/assert removal after refactor | Keep |
| `user-agent-data-types` | devDependency | 2023-09-08 | Type support for `navigator.userAgentData` | None in audit for this package | None | Local `.d.ts` | Keep |
| `scroll-into-view-if-needed` | dependency | 2023-09-13 | Message scroll behavior in chat | None in audit for this package | None | Native `scrollIntoView` | Keep |
| `js-md5` | dependency | 2023-10-09 | Identicon hash helper | None in audit for this package | None | `crypto.subtle.digest` | Keep |
| `@rollup/plugin-inject` | devDependency | 2023-10-15 | Injects `Buffer` in build pipeline | None in audit for this package | None | Remove after polyfill cleanup | Keep |
| `@mdi/js` | dependency | 2023-12-27 | Icon paths across UI components | None in audit for this package | None | Other icon sets | Keep |

## Proposed execution plan
1. PR 1: remove clearly unused dependencies (`simple-audio`, `i`, `jest-serializer-vue`, `@esbuild-plugins/node-globals-polyfill`) and validate build/test
2. PR 2: replace low-risk utility packages (`copy-to-clipboard`, `b64-to-blob`, `file-saver`, `visibilityjs`, `notifyjs`)
3. PR 3: queue and state modernization (`promise-queue`, `throttle-promise`, `vuex-persist`) with regression tests
4. PR 4: long-track store migration (`vuex` to Pinia) scoped by modules and feature flags where required
5. PR 5: crypto derivation migration (`hdkey` + possible `bip39` review) with deterministic test vectors and cross-wallet compatibility checks
6. PR 6: optional browser polyfill cleanup (`https-browserify`, `os-browserify`, `stream-*`, `path-browserify`, `rollup-plugin-node-polyfills`, `@rollup/plugin-inject`) only after dependency graph reduction

## Acceptance criteria
- No regressions in auth, key derivation, transaction signing, message encryption/decryption, or pending transaction behavior
- No regressions in PWA, Electron, and Capacitor build targets
- `npm run lint`, `npm run typecheck`, `npm run test -- --run`, and relevant build commands pass
- Dependency count is reduced and stale utility packages are replaced with maintained equivalents

## Additional notes
- `vuedraggable` has npm dist-tag confusion (`npm view vuedraggable version` shows `2.24.3`) but installed `4.1.0` was published on 2021-08-22 and is used in Wallets view
- This task intentionally separates dependency hygiene from protocol-level cryptography changes to reduce risk

### LINKED ISSUE #958: [Composite, Task] Full security audit and coordinated hardening release
### Summary

Complete a full, code-level security audit of the PWA across all four targets (Web/PWA, Tor, Electron desktop, Capacitor Android) and land the resulting hardening work as one coordinated release.

### Details

A fresh audit of the current `dev` tree has been completed. It reviews the application source, build and deployment configuration, platform manifests and the dependency tree, and it consolidates all earlier security material into a single plan.

Scope of the audit:

- application source in `src/**` for Web/PWA, Tor, Electron and Capacitor Android targets
- build, deployment and platform configuration (`vite-*.config.ts`, `vercel.json`, `api/`, `android/`, `capacitor.config.ts`, `.github/workflows/`)
- dependency tree as resolved by `package-lock.json`

Out of scope for this repository:

- `adamant-node`, `ipfs-node` and other backend services
- host and nginx configuration for the official web endpoints
- blockchain consensus and the strength of the NaCl/Ed25519 primitives themselves

Methods applied:

- manual review of the cryptography, rendering, storage, node-transport and platform layers
- targeted dynamic probing of the rendering pipeline through Vitest with jsdom, with captured output
- dependency analysis via `npm audit` against the current lockfile
- differential review of #952 to identify defects and regressions before reusing any of its work

Findings are grouped into six areas: message rendering and sanitization, local secret handling and key derivation, trust in single-node responses, desktop and mobile platform hardening, HTTP security headers across deployment hosts, and supply chain plus CI gating.

Technical details, individual findings and proof-of-concept material are deliberately **not** published here. They stay internal until the fix release has shipped and users have had time to update, in line with responsible disclosure. The working documents live in `.ai-tasks/` on the working branch.

Cryptographic core assessment: signing, message encryption and key derivation are implemented correctly. The work in this issue is about the surrounding layers, not about replacing primitives.

The audit also re-validated the earlier community findings against the current tree and formally closed out several of them as false positives or accepted risks with written rationale, so the outcome is documented either way.

### Checklist

- [ ] Phase 0 — preparation: logging helpers, shared test fixtures for the rendering pipeline
- [ ] Phase 1 — rendering and sanitization pipeline, plus a lint rule to prevent regressions
- [ ] Phase 2 — local secret handling and key derivation, with a tested migration path for existing users
- [ ] Phase 3 — node trust model: cross-node verification, realtime event handling, download bounds
- [ ] Phase 4 — platform hardening: Electron shell, Android manifest and providers, HTTP security headers, transport policy
- [ ] Phase 5 — supply chain remediation and CI security gates (audit, SAST, Dependabot)
- [ ] Phase 6 — remaining code quality items, `AGENTS.md` policy updates, release notes
- [ ] Cross-platform manual verification pass (Web, Tor, Electron, Android)
- [ ] Follow-up issues filed for the items that must be fixed outside this repository

### Notes

Related work:

- #952 — earlier hardening attempt, still open against `dev`. Its approach is broadly sound and parts of it will be reused, but it cannot be merged as-is: the differential review found defects and platform regressions that must be corrected first, and it mixes unrelated cleanups with the security changes. It will be superseded by the phased work tracked here.
- #929 — stale direct dependency review and phased replacement plan (overlaps with Phase 5)
- #927 — remove runtime `unsafe-eval` sources and resume strict CSP hardening (follow-up to Phase 4; nonce-based CSP stays out of this release)

Three items require action outside this repository and will be filed separately with their owners:

- a node-side change in `adamant-node`
- host configuration for the official web endpoints
- GitHub repository security settings (Dependabot security updates, secret scanning, code scanning)

Decisions needed from maintainers before implementation starts:

- session-restore behaviour after the secret-handling change, which has a visible UX effect
- node list and transport policy tradeoffs, since the project must stay usable under degraded connectivity and censorship
- the blocking threshold for the dependency audit gate in CI, and which residual advisories are formally accepted
- Android backup policy
- how much markdown rendering fidelity to preserve under a stricter allowlist
- disclosure timing for the audit document

Working principle for the whole programme: each change stays minimal and targeted, every phase carries regression tests near the changed behaviour, and no phase weakens decentralization, censorship resistance or offline tolerance in order to gain hardening.

### Verification

Automated, after every phase and before release:

- `npm run lint`
- `npm run typecheck`
- `npm run test -- --run`
- `npm run build`
- `npm run build:tor`
- `npm run build:testnet`
- `npm run electron:build`
- `npm run android:prebuild`
- `npm run test:e2e`
- `npm audit signatures`

Manual, on each platform (Web, Tor, Electron, Android):

- message rendering behaves correctly for hostile input in the message body, chat list, reply preview and quoted message, with message formatting both enabled and disabled
- a local database written before the record-format change is **not** migrated, by decision: it fails to decrypt, is cleared as a whole, Stay logged in is turned off and the user signs in with the passphrase once. Only cached blockchain data is lost, and it is rebuilt from the chain. Incorrect credentials fail the same way, and no store is left partially written. A one-release legacy read path was designed and rejected — at the current user count the temporary compatibility code costs more than the data it would preserve, and such code tends to outlive its release
- messaging continues to work with reduced node connectivity, and surfaces a clear error when a security check does not pass
- desktop file access boundaries and external link handling behave as intended
- attachment download limits are enforced before decryption
- security headers are present on every official deployment host
- no regression in dark theme, notifications, attachments, transfers or node switching

Definition of done: all phases merged into `dev` through reviewable PRs, cross-platform manual pass completed, release notes published covering the user-visible behaviour changes, follow-up issues filed for out-of-repository items, and the audit document published only after the fix release has been adopted.

al-onyxprotocol:
## Description

This is the second security-hardening batch from #958, based on the full audit in the branch-local `.ai-tasks` documents. It focuses on local secret storage and password KDF correctness, dependency and CI hardening, and strict CSP enforcement. The previously merged rendering and node-trust work is not duplicated here.

### Local secrets and password KDF

- Replaces the legacy password hash path with a versioned scrypt descriptor (`N=2^15`, `r=8`, `p=1`) and fixes the secret-key byte conversion that previously reduced effective entropy
- Keeps the measured scrypt working set at approximately 32 MiB while allowing 64 MiB through `maxmem` for implementation overhead
- Runs scrypt in a worker with a 30-second timeout and a lazy synchronous fallback
- Removes passphrases and secret-bearing stores from `window`, `sessionStorage`, Electron `executeJavaScript`, and the renderer/main bridge
- Makes IndexedDB writes await transaction completion and clears an incompatible legacy cache fail-safe instead of partially restoring it
- Adds regression coverage for password unlock after a real reload, legacy-cache cleanup, stable descriptors, malformed data, and secret exposure contracts

A full reload now requires credentials again in both login modes: a passphrase-only session returns to full passphrase login, while a password-protected session shows the password unlock form. A legacy encrypted cache without a KDF descriptor is cleared once and requires passphrase login; cached chain data is rebuilt afterwards.

### Dependencies, runtime, and CI

- Migrates cryptographic helpers to `@noble/hashes`, `@scure/bip32`, and `@scure/bip39` with deterministic ADM, BIP39, ETH, BTC-family, and identicon vectors
- Removes obsolete wrappers, duplicate crypto stacks, unused packages, and unused Node browser polyfills
- Migrates to ESLint 10 through `eslint-plugin-import-x` while preserving import correctness probes; this completes #928
- Pins Node 24 and npm 12 and activates the pinned npm through Corepack in every repository workflow that executes npm
- Adds Dependabot and audit/signature security gates, with the remaining transitive advisory documented and time-bounded in `DEPENDENCY_SECURITY.md`
- Runs a production PWA build in the mandatory quality workflow so CSP injection, generated chunks, the service worker, and the CSP bundle gate are checked on every PR
- Splits live no-broadcast wallet setup by coin so one unavailable indexer no longer suppresses unrelated wallet coverage

Direct dependencies decrease from 78 to 67 and direct dev dependencies from 63 to 55. The installed tree is 1,361 packages; the production audit has zero findings, while the full audit has three accepted moderate findings in the Capacitor CLI toolchain.

### Strict CSP

- Removes all direct `eval` and `Function` constructor sources from generated JavaScript
- Replaces the string-capable `setImmediate` implementation with a callback-only local polyfill
- Removes `lodash-es` from browser runtime modules and keeps it only as a build-time wallet-generator dependency
- Rewrites the exact `engine.io-client` global fallback to `globalThis` during bundling
- Injects an enforced CSP into PWA, Tor, testnet, Android, and Electron static artifacts and aligns the Vercel response header
- Adds a build gate that rejects any generated chunk containing a direct `eval` or `Function` constructor
- Hardens the CSP-report endpoint for raw `application/csp-report` bodies, exact Vercel host boundaries, bounded payloads, and enforcing-policy event labels

The policy deliberately moves in two directions. `script-src` becomes substantially narrower: it removes inline and JavaScript string execution and all remote script hosts. `connect-src`, `img-src`, and `media-src` allow arbitrary HTTP(S) and WebSocket origins so user-configured and self-hosted nodes keep working. The single script-policy exception is `wasm-unsafe-eval`, required by the bundled secp256k1 WebAssembly implementation; it permits WebAssembly compilation without re-enabling JavaScript string evaluation.

The generated-bundle gate scans reachable emitted chunks and direct lexical calls. It is a guard against accidental reintroduction, while runtime CSP remains the enforcement boundary for indirect aliases and computed access. Meta-delivered CSP cannot enforce `frame-ancestors`; Vercel and Electron enforce it through response headers, and the remaining deployed-host matrix stays a post-merge verification step in #927.

### Theme behavior

On a fresh profile with no saved preference, the app uses the operating-system theme at startup. After the user selects a theme, `darkTheme` remains persisted in local storage and wins on later launches. Subsequent OS theme changes do not overwrite that explicit choice because the UI has no separate "Follow system" mode.

### Measured performance

The perceived startup improvement is reproducible. Compared with `dev` at `cbfa83db`, the same local production build and browser conditions produced:

| Metric | `dev` | This branch | Change |
|---|---:|---:|---:|
| Main JavaScript, gzip | 1,194.71 kB | 762.64 kB | -36.2% |
| CSS, gzip | 131.03 kB | 97.91 kB | -25.3% |
| Transformed modules | 2,981 | 2,398 | -19.6% |
| Vite production-build phase | 20.57 s | 7.10 s | -65.5% |
| Login form ready, median | 3,642 ms | 2,924 ms | -19.7% |
| DOMContentLoaded, median | 1,595 ms | 1,195 ms | -25.1% |

The browser comparison used seven cold-cache Chromium runs per revision with service workers cleared, 40 ms latency, 10 Mbps throughput, and 4x CPU throttling. Transfer size decreased by 12.8% and decoded body size by 22.2%. These figures measure initial local rendering, not remote node readiness.

### Deliberately not included

TypeScript 7 was evaluated but not adopted. The native compiler is substantially faster in a raw TypeScript-only probe, but the current Vue language tooling cannot execute against it, and `typescript-eslint` currently supports TypeScript only below 6.1. The main `vue-tsc` path therefore still needs TypeScript 6; adding a second compiler would add CI and editor complexity for little project-level benefit.

## Related issues

- Part of #958
- Implements the code-side work for #927
- Closes #928
- Advances #929

## How to test

Run with Node 24 and the pinned npm 12 release:

```bash
corepack npm ci
corepack npm run lint
corepack npm run typecheck
corepack npm run test -- --run
corepack npm run build
corepack npm run build:testnet
corepack npm run build:tor
corepack npm run electron:build
corepack npm run android:prebuild
corepack npm run test:e2e
corepack npm audit --omit=dev --audit-level=high
corepack npm audit --audit-level=high
corepack npm audit signatures
```

Manually verify a password-protected reload in packaged Electron and Tor builds. After merge, verify the CSP header and core flows on the deployment matrix listed in #927.

## Verification performed

- Node 24.19.0 and npm 12.0.2
- ESLint: clean with zero warnings
- Typecheck: clean (`vue-tsc` and Node/Vite configs)
- Unit tests: 135 files, 829 tests passed
- Default Playwright matrix: 108 passed in CI; one known chat-composer navigation flake passed on retry
- Long-running BTC pending-transaction visual test: passed
- Live no-broadcast wallet builders: 26 tests passed; no non-ADM transfer was broadcast
- PWA, testnet, Tor, Android prebuild, and Electron renderer/main/preload builds: passed
- Universal packaged Electron build: passed for x64 and arm64
- Generated PWA, Android, and Electron JavaScript: no direct `eval` or `Function` constructors
- Vercel preview: enforced CSP header verified; password reload/legacy-cache and chats-core flows passed in Chromium
- Vercel preview CSP report: a deliberately blocked external script produced an `application/csp-report` request with `disposition: enforce`; the endpoint returned 204
- Surge preview: injected strict CSP meta verified
- Production dependency audit: zero findings
- Full dependency audit: three documented moderate findings, zero high or critical findings
- Package signatures: 1,361 verified packages and 365 attestations

## Checklist

- [x] Security and cryptographic compatibility covered by deterministic vectors
- [x] Reload and legacy-storage behavior covered by E2E tests
- [x] Runtime dependency and generated-bundle CSP gates added
- [x] PWA, Tor, Electron, and Android build paths verified
- [x] No passphrase, private key, plaintext payload, or token logging introduced
- [x] No analytics, tracking, or centralized endpoint introduced
- [ ] Post-merge deployed-host CSP matrix verified


.github/workflows/eslint.yml
```diff
@@ -19,17 +19,24 @@ env:
...
+        run: npm run lint
+
+      - name: Verify ESLint produced no changes
+        run: git diff --exit-code
+
+      - name: Run typecheck
+        run: npm run typecheck
+
+      - name: Run unit tests
+        run: npm run test -- --run
```
adamant-al, line 42: **Medium — CI-Lücke**

Der Quality-Job endet bei Lint, Typecheck und Unit-Tests. Ein Production-`npm run build` fehlt.

Damit werden auf dem PR **nicht** geprüft:
- injiziertes CSP-Meta
- das `generateBundle`-Gate gegen `eval` / `Function`
- Service-Worker- / Chunk-Artefakte

Playwright-Smoke bedient typischerweise den Dev-Server und schließt diese Lücke nicht.

**Vorschlag:** einen Build-Schritt (zumindest Web-PWA) in Quality oder Security aufnehmen, damit das CSP-Gate vor dem Merge grün sein muss.

src/store/plugins/sessionStorage.js
```diff
@@ -1,17 +1,50 @@
 import VuexPersistence from 'vuex-persist'
 
+export function reduceSessionState(state) {
+  return {
+    // Deliberately exclude the passphrase and password hash. A refresh must require the user
+    // to unlock the encrypted cache again instead of leaving account secrets in Web Storage.
+    address: state.address,
+    balance: state.balance
+  }
```
adamant-al, line 9: **Low — UX / Dokumentation**

Richtig und gewollt: weder Passphrase noch Password-Hash in `sessionStorage`.

Nebenwirkung: nach einem **vollen Reload** müssen sich auch Nutzer **ohne** „Stay logged in“ (reine Passphrase-Session) erneut anmelden — `isLogged` hängt an `passphrase.length`, und die Passphrase liegt nicht mehr in der Session.

Die PR-Beschreibung betont vor allem den Password-Unlock-Pfad. Bitte explizit ergänzen, dass **beide** Login-Modi nach Full Reload eine erneute Eingabe brauchen — sonst wirkt F5 wie ein Regression-Bug.

adamant-al (commented):
## Review-Zusammenfassung (Security Hardening Part 2)

**Verdict:** Comment — keine blockierenden Security-Findings im Diff. Starker Hardening-Batch; ein CI-Gap und dokumentarische/Rest-Risiken sollten bewusst bleiben.

### Was gut ist
- **Password-KDF:** versioniertes scrypt (`N=2^15`, `r=8`, `p=1`), CSPRNG-Salt, strikte Descriptor-Validierung (Anti-DoS), Worker + Timeout + Fallback
- **IDB-Crypto:** `hexToBytes` vor `ed2curve` schließt den alten Entropy-Verlust; frischer Nonce pro Record (bereits Layout `nonce || ciphertext`); Legacy bewusst inkompatibel
- **Secret exposure:** kein `window.store` / `window.ep`; Session speichert keine Passphrase/Password-Hashes; Electron-Preload nur `isElectron`
- **CSP:** kein JS-`unsafe-eval`; Build-Gate; engine.io-Patch; `setImmediate` nur mit Callbacks
- **Deps/CI:** `@noble`/`@scure`, weniger Direct Deps, Dependabot, Audit-Signatures, CodeQL, gepinntes Node 24 / npm 12
- **Tests:** Unit-Verträge + E2E Password-Reload / Legacy-Wipe

### Worauf achten
1. **Quality-CI baut nicht production** — CSP-/`eval`-Gate läuft nicht auf jedem PR
2. **UX:** Full Reload verlangt Re-Login auch für reine Passphrase-Sessions (PR-Text nennt bisher vor allem Password-Mode)
3. **CSP-Trade-offs:** `'wasm-unsafe-eval'`, `style-src 'unsafe-inline'`, weites `connect-src`/`img-src` für Custom Nodes
4. **Post-merge:** Host-Matrix aus #927 (Meta vs. HTTP-CSP, `frame-ancestors` auf Surge/Pages)
5. **Backlog außerhalb dieses PR:** ADM-SEC-026 / ADM-SEC-027 laut Audit-Checklist

### Findings-Übersicht

| Severity | Ort | Finding |
|----------|-----|---------|
| Medium | `.github/workflows/eslint.yml` | Quality-Job ohne `npm run build` — CSP-Bundle-Gate nicht PR-seitig erzwungen |
| Low | `sessionStorage.js` / PR-Beschreibung | Passphrase-Sessions brauchen nach Reload ebenfalls Login; in der PR-Beschreibung unterbetont |
| Note | CSP (`connect-src` / `img-src`) | Bewusst weit für Custom Nodes; bei XSS erleichtert Exfiltration (XSS hat ohnehin Speicherzugriff) |
| Note | `script-src 'wasm-unsafe-eval'` | Residual für secp256k1-WASM; JS-`eval` ist entfernt |
| Note | CSP Build-Gate | Erkennt nur unqualifiziertes `eval(` / `Function(`; Runtime-CSP bleibt die echte Kontrolle |
| Note | `package.json` engines | Node ≥24.15 / npm ≥12 — Breaking für Contributor auf 20/22 |
| Note | Electron Linux `--no-sandbox` | Pre-existing Trade-off, nicht neu eingeführt |

Inline-Kommentare markieren die konkreten Stellen (Findings und positive Beobachtungen).

vite-config/plugins/cspHardeningPlugin.ts
```diff
@@ -0,0 +1,110 @@
+import type { Plugin } from 'vite'
+
+export const PWA_CONTENT_SECURITY_POLICY = [
+  `default-src 'self'`,
+  `base-uri 'self'`,
+  `object-src 'none'`,
+  `frame-src 'none'`,
+  `script-src 'self' 'wasm-unsafe-eval'`,
```
adamant-al, line 10: **Note — Residualrisiko `wasm-unsafe-eval`**

Erlaubt WASM-Compilation (secp256k1), **nicht** JavaScript-`eval`. Das ist der schmalste praktikable Ausnahme-Pfad.

Rest-Risiko: bei bestehendem XSS kann ein Angreifer zusätzlich WASM laden/compilieren. Dominantes Risiko bleibt dann ohnehin der JS-Kontext — kein Blocker, aber bewusst dokumentieren (passt zu #927 / Audit-Backlog).

vite-config/plugins/cspHardeningPlugin.ts
```diff
@@ -0,0 +1,110 @@
...
+  `style-src 'self' data: 'unsafe-inline'`,
+  `img-src 'self' data: blob: http: https:`,
+  `media-src 'self' data: blob: http: https:`,
+  `font-src 'self' data:`,
+  `connect-src 'self' http: https: ws: wss: blob:`,
```
adamant-al, line 16: **Note — bewusst weites `connect-src`**

`http:` / `https:` / `ws:` / `wss:` sind weiter als der frühere Domain-Allowlist und nötig für **user-konfigurierbare Nodes** (Dezentralisierung).

Trade-off: bei XSS ist Exfiltration zu beliebigen Endpunkten CSP-seitig leichter. Sobald XSS existiert, liegen Secrets ohnehin im Speicher — deshalb kein Medium-Finding, aber der Trade-off sollte in der Security-Doku/PR sichtbar bleiben.

Gleiches Muster bei `img-src` / `media-src` darüber.

vite-config/plugins/cspHardeningPlugin.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  if (hasUnqualifiedCall(code, 'Function')) {
+    violations.push(`${fileName}: Function constructor`)
+  }
+
+  if (hasUnqualifiedCall(code, 'eval')) {
+    violations.push(`${fileName}: direct eval`)
+  }
+
+  return violations
```
adamant-al, line 71: **Note — Build-Gate-Abdeckung**

`hasUnqualifiedCall` erkennt `eval(` / `Function(`, aber **nicht** z. B. `globalThis.eval(` oder `window['Function'](...)`.

Das ist okay als Defense-in-Depth: die **Runtime-CSP** ohne `'unsafe-eval'` blockiert die Ausführung weiterhin. Nur nicht erwarten, dass das Gate jede indirekte Form statisch findet.

vercel.json
```diff
@@ -11,7 +11,7 @@
...
+          "value": "default-src 'self'; base-uri 'self'; object-src 'none'; frame-ancestors 'none'; frame-src 'none'; script-src 'self' 'wasm-unsafe-eval'; style-src 'self' data: 'unsafe-inline'; img-src 'self' data: blob: http: https:; media-src 'self' data: blob: http: https:; font-src 'self' data:; connect-src 'self' http: https: ws: wss: blob:; worker-src 'self' blob:; manifest-src 'self'; form-action 'self'; report-uri /api/csp-report;"
```
adamant-al, line 14: **Note — HTTP-CSP vs. Meta-CSP**

Positiv: Header-CSP mit `frame-ancestors 'none'` und `report-uri /api/csp-report` (Endpoint existiert unter `api/csp-report.js`).

Meta-CSP (PWA-Plugin) kann `frame-ancestors` in Browsern **nicht** durchsetzen — deshalb ist der Vercel-Header wichtig. Hosts ohne Response-Header (Surge / GitHub Pages / DeWeb) brauchen die Post-Merge-Matrix aus #927 bzw. ggf. `X-Frame-Options`, wo die Plattform das hergibt.

package.json
```diff
@@ -14,10 +14,10 @@
     "type": "git",
     "url": "git+https://github.com/Adamant-im/adamant-im.git"
   },
-  "packageManager": "npm@10.9.4",
+  "packageManager": "npm@12.0.2",
   "engines": {
-    "node": "^20.19.0 || >=22.12.0",
-    "npm": ">=10"
+    "node": ">=24.15.0",
+    "npm": ">=12"
```
adamant-al, line 20: **Note — Breaking Engine-Bump**

`node: >=24.15.0` und `npm: >=12` (plus `.nvmrc` v24 / Corepack in Workflows) sind konsistent mit dem Hardening, aber **breaking** für Contributor und lokale Toolchains auf Node 20/22.

README/Contributing kurz erwähnen, falls noch nicht überall nachgezogen.

src/lib/idb/passwordKdf.ts
```diff
@@ -0,0 +1,218 @@
...
+  return (
+    descriptor.v === PASSWORD_KDF_PARAMETERS.v &&
+    descriptor.alg === PASSWORD_KDF_PARAMETERS.alg &&
+    descriptor.N === PASSWORD_KDF_PARAMETERS.N &&
+    descriptor.r === PASSWORD_KDF_PARAMETERS.r &&
+    descriptor.p === PASSWORD_KDF_PARAMETERS.p &&
+    descriptor.dkLen === PASSWORD_KDF_PARAMETERS.dkLen &&
+    typeof descriptor.salt === 'string' &&
+    SALT_HEX_PATTERN.test(descriptor.salt)
+  )
```
adamant-al, line 74: **Gut — Descriptor-Pinning**

Nur exakt die festgelegten Parameter (`N`/`r`/`p`/`dkLen`/`alg`/`v`) plus hex-Salt werden akzeptiert. Verhindert, dass ein manipuliertes `localStorage`-Descriptor mit riesigem `N` zum DoS beim Unlock wird. Sauber.

src/lib/idb/passwordKdf.ts
```diff
@@ -0,0 +1,218 @@
...
+        finish(() => reject(new Error(message)))
+      }
+    }
+
+    worker.onerror = (event) => {
+      event.preventDefault()
+      useFallback()
+    }
+
+    timeout = setTimeout(useFallback, timeoutMs)
```
adamant-al, line 208: **Gut — Worker-Timeout + Fallback**

Nach 30 s wird der Worker terminiert und derselbe scrypt-Pfad synchron/lazy nachgeladen. `settled` verhindert Double-Resolve. Gute Balance aus UI-Responsiveness und Krypto-Korrektheit.

src/lib/idb/crypto.js
```diff
@@ -19,7 +18,19 @@ function getPasswordHash(password) {
...
 
 function getSecretKey() {
-  return ed2curve.convertSecretKey(getPasswordHash(store.state.password))
+  const passwordHash = getPasswordHash(store.state.password)
+
+  if (typeof passwordHash !== 'string' || !/^[0-9a-f]{64}$/.test(passwordHash)) {
+    throw new Error('Invalid password hash')
+  }
+
+  const secretKey = ed2curve.convertSecretKey(hexToBytes(passwordHash))
```
adamant-al, line 27: **Gut — ADM-SEC-025 Fix**

`hexToBytes(passwordHash)` vor `ed2curve.convertSecretKey` stellt sicher, dass die vollen 32 Byte des scrypt-Outputs ankommen — nicht mehr der verlustbehaftete String-Pfad, bei dem Hex-Buchstaben faktisch keine Entropie trugen.

src/lib/idb/crypto.js
```diff
@@ -19,7 +18,19 @@ function getPasswordHash(password) {
...
+
+  if (!secretKey) {
+    throw new Error('Failed to derive the IDB encryption key')
+  }
+
+  return secretKey
 }
 
 /**
```
adamant-al, line 36: **Gut — Per-Record-Nonce / Legacy-Wipe**

Layout `nonce || ciphertext` und bewusste Inkompatibilität zum Shared-Nonce-Format sind die richtige Abwägung: Shared-Keystream/XOR und kaputte Poly1305-Auth sind schlimmer als einmaliger Cache-Verlust + Passphrase-Login.

(Der Nonce-Pfad selbst war schon vor diesem Diff im Layout; dieser PR härtet den Key-Derivation-Zweig und den Wipe-Pfad dazu.)

src/lib/idb/crypto.js
```diff
@@ -62,24 +73,9 @@ export function decrypt(encryptedData) {
...
-      (err, encryptedPassword) => {
-        if (err) return reject(err)
-
-        const passwordString = bytesToHex(encryptedPassword)
-
-        resolve(passwordString)
-      }
-    )
-  })
+export function encryptPassword(password, descriptor) {
```
adamant-al, line 79: **Gut — PBKDF2 → scrypt-Deskriptor**

`encryptPassword` delegiert nur noch an `derivePasswordHash(password, descriptor)`. Kein festes Salt/Iteration-Set mehr im Client-Konstantenblock — Salt ist pro User im Descriptor.

src/store/plugins/indexedDb.js
```diff
@@ -136,26 +173,10 @@ export default (store) => {
...
-              store.commit('reset')
-            })
-            .catch((err) => {
-              logger.log('indexed-db-plugin', 'warn', err)
-            })
-            .finally(() => {
-              router.push('/')
-            })
         })
     }
```
adamant-al, line 181: **Note — Reload ohne In-Memory-Password**

Wenn `stayLoggedIn` + gültiger KDF-Descriptor gesetzt sind, aber `store.state.password` nach Reload fehlt (nicht mehr in der Session), greift weder Restore noch Fallback — die UI zeigt Unlock. Das ist korrekt und durch E2E abgedeckt.

Legacy ohne Descriptor → `fallBackToPassphraseLogin` + `clearDb()` darüber: ebenfalls gut.

src/lib/idb/stores/Security.js
```diff
@@ -32,7 +32,7 @@ export default {
         })
       })
 
-      return tx.complete
+      return tx.done
```
adamant-al, line 35: **Gut — `tx.done` statt `tx.complete`**

Unter `idb` v8 ist `tx.done` das Completion-Promise. `tx.complete` war faktisch wirkungslos — Writes konnten als „fertig“ gelten, bevor die Transaktion committed war. Wichtiger Korrektheits-Fix (Contract-Test vorhanden).

src/electron/main.js
```diff
@@ -203,6 +189,7 @@ function createWindow() {
     // These are the current Electron defaults. They are stated explicitly so that a future
     // Electron upgrade, or a copy of this config, cannot silently relax them.
     webPreferences: {
+      preload: path.join(__dirname, 'preload.cjs'),
```
adamant-al, line 192: **Gut — Electron Renderer-Isolation**

`preload.cjs` + `contextIsolation` + `sandbox` + `nodeIntegration: false` explizit gesetzt. Dazu: kein `executeJavaScript` mehr gegen den Renderer, Preload exponiert nur `isElectron`.

**Note daneben:** Die Electron-CSP (`electronRendererCsp`) folgt dem PWA-Muster (`'wasm-unsafe-eval'`, weites `connect-src`). `frame-ancestors 'none'` wirkt hier über den Response-Header der `app://`-HTML-Antworten — im Gegensatz zu Meta-only auf manchen Static Hosts.

src/electron/preload.ts
```diff
@@ -0,0 +1,8 @@
+import { contextBridge } from 'electron'
+
+contextBridge.exposeInMainWorld(
+  'adamantDesktop',
+  Object.freeze({
+    isElectron: true as const
+  })
```
adamant-al, line 7: **Gut — minimales Preload**

Nur ein eingefrorenes `isElectron`-Flag über `contextBridge`. Keine Secret-/Store-Bridge. Contract-Test in `secretExposureContract.spec.ts` deckt das ab.

src/store/index.js
```diff
@@ -286,7 +300,6 @@ const store = {
 }
 
 const storeInstance = createStore(store)
```
adamant-al, line 302: **Gut — kein `window.store` mehr**

Die globale Store-Exposition ist entfernt (analog `window.ep` in `main.ts`). Verringert die Angriffsfläche für XSS/Extensions, die zuvor triviale Store-Enumeration hatten. Secrets bleiben trotzdem im JS-Heap, solange die Session offen ist — das ist unvermeidbar, aber die DOM-Globalen waren unnötig.

src/main.ts
```diff
@@ -36,6 +34,4 @@ registerGlobalComponents(app)
 
 app.mount('#app')
```
adamant-al, line 35: **Gut — kein `window.ep` mehr**

Die Vue-App-Instanz wird nicht mehr am `window` aufgehängt. Zusammen mit dem entfernten `window.store` schließt das die offensichtlichen Secret-/State-Globals aus dem früheren Debug-Pfad.

DEPENDENCY_SECURITY.md
```diff
@@ -0,0 +1,24 @@
...
+The runtime dependency tree has no known advisories as of 2026-08-10. The full development tree has
+one accepted moderate advisory chain:
+
+- `@capacitor/cli -> xcode -> uuid@7` (`GHSA-w5hq-g745-h8pq`). `xcode` is a development-only iOS
+  project editor, while this repository's Capacitor target is Android. The package calls UUID v4;
+  the advisory affects caller-provided buffers in UUID v3, v5 and v6. No supported Capacitor 8.5
+  release currently removes this transitive version.
+
+This acceptance expires on 2026-11-10 and must be reviewed earlier when Capacitor publishes an
+update. A newly introduced high or critical advisory is never covered by this acceptance and is
```
adamant-al, line 19: **Gut — zeitlich begrenzte Ausnahme**

Capacitor/`uuid@7`-Kette klar als **dev-only**, Advisory-Scope und **Expiry 2026-11-10** dokumentiert. High/Critical bleiben CI-Blocker. Bitte Reminder setzen, vor Ablauf erneut zu prüfen, sobald Capacitor nachzieht.

src/lib/cloneState.ts
```diff
@@ -0,0 +1,8 @@
+import { toRaw } from 'vue'
+
+/**
+ * Clone Vuex state without retaining Vue proxies or references to mutable nested values.
+ */
+export function cloneState<T>(value: T): T {
+  return structuredClone(toRaw(value))
```
metalisk, line 7: `structuredClone` is not a drop-in replacement for `cloneDeep` — it changes the failure mode from silent to throwing. I checked the three cases on Node 24:

- `structuredClone(new Proxy({a:1}, {}))` → **throws** `DataCloneError`
- an object holding a function → **throws** `DataCloneError`
- a class instance → clones, but the prototype is dropped (`Object.getPrototypeOf(clone) === Object.prototype`, methods gone)

`toRaw()` only unwraps the root proxy, so this is safe as long as nothing ever stores a reactive proxy *inside* the state, and as long as every value is structured-cloneable.

Today `state.chat` holds plain data and everything ends up in `JSON.stringify` anyway, so nothing is actually lost that was not already lost. The difference is what happens when that stops being true: `cloneDeep` degraded quietly, this throws. See the call site for where the throw lands.

Worth a regression test that puts a non-cloneable value into the state and asserts the intended behaviour — whichever you decide it should be.

src/lib/idb/state.js
```diff
@@ -40,7 +40,7 @@ function cloneModules(state) {
 
   // clone `chat` module, except `chats` key
   if (state.chat) {
-    const chat = cloneDeep(state.chat)
+    const chat = cloneState(state.chat)
```
metalisk: Follow-up to the `cloneState` note: this call is **synchronous and outside the promise chain** in `saveState`, so a `DataCloneError` propagates as a synchronous throw rather than a rejected promise.

`saveState` has a single production caller, `PasswordSetDialog`, where it runs inside a `.then()` — so the throw becomes a rejected promise and lands in the existing `.catch`, which calls `clearDb()` and `removePassword()`.

Net effect if a non-cloneable value ever appears in `state.chat`: the user does not see an error, they silently lose "Stay logged in" and the encrypted cache. That is the fail-safe working as designed, but it makes the cause invisible. A narrow `try/catch` here that logs the failing module name would turn a mystery into a one-line log.

vite-config/plugins/cspHardeningPlugin.ts
```diff
@@ -0,0 +1,110 @@
...
+  if (!ENGINE_IO_GLOBALS_MODULE.test(id)) return null
+
+  const hardenedCode = code.replace(ENGINE_IO_GLOBAL_FALLBACK, 'globalThis')
+  return hardenedCode === code ? null : hardenedCode
+}
+
+export function findUnsafeRuntimeSources(fileName: string, code: string): string[] {
+  const violations = []
+
+  if (hasUnqualifiedCall(code, 'Function')) {
```
metalisk, line 63: I verified this gate actually fires rather than taking it on trust: adding `new Function('return 1')` inside a live code path (`bytesToHex`) fails the build with `Strict CSP build gate rejected runtime code: assets/…: Function constructor`. Good.

Two boundaries worth writing down next to it, because both surprised me while testing:

1. **Only emitted chunks are scanned.** My first probe was an exported-but-unused `__cspProbe`, and the build passed — Rolldown tree-shook it before `generateBundle`. That is correct behaviour (code that is not in the bundle cannot violate the policy), but it means "the gate passed" says nothing about source that is not reachable.
2. **Detection is lexical.** `hasUnqualifiedCall` looks for `Function(` not preceded by an identifier character or `.`, so `const F = Function; F('…')`, `globalThis['Function']` and `[]['constructor']['constructor']` all pass. Minifiers cannot rename a global, so ordinary regressions are caught — but this is a guard against accidental reintroduction, not proof of absence.

The README already frames it as a build gate rather than a proof, which is the right framing; a comment here would keep that clear for whoever touches this next.

vite-config/plugins/cspHardeningPlugin.ts
metalisk, line 16: Worth naming explicitly in the PR description, because the policy moves in **two directions at once**:

- **Tightened, significantly**: `script-src` drops `'unsafe-inline'` and `'unsafe-eval'` and the `*.adamant.im` / `*.adm.im` / `*.bbry.org` host allowances, leaving `'self' 'wasm-unsafe-eval'`.
- **Loosened**: `connect-src`, `img-src` and `media-src` go from that same domain allowlist to `http:` / `https:` / `ws:` / `wss:` — i.e. any origin.

The loosening is justified — users configure their own nodes, and pinning to ADAMANT domains would break that — and with `script-src 'self'` and no inline execution, getting code into the page to abuse the wide `connect-src` is now much harder. So I think the trade is right.

But a reader of the diff sees "CSP hardening" and will not expect three directives to get broader. Stating it in the description also gives the next reviewer the reason, so nobody narrows `connect-src` later and breaks custom nodes.

vite-config/plugins/cspHardeningPlugin.ts
```diff
@@ -0,0 +1,110 @@
+import type { Plugin } from 'vite'
+
+export const PWA_CONTENT_SECURITY_POLICY = [
+  `default-src 'self'`,
+  `base-uri 'self'`,
+  `object-src 'none'`,
+  `frame-src 'none'`,
```
metalisk, line 9: `frame-ancestors` is absent here, and it cannot be added: the directive is ignored when a policy is delivered through a `<meta>` tag, it only works as a response header.

On Vercel that is covered — `vercel.json` sets both `frame-ancestors 'none'` and `X-Frame-Options`. The gap is the other static targets that get only this meta policy: GitHub Pages, the deweb deployment and the Tor build have no framing protection from CSP.

Not something this plugin can fix, but worth a comment here so the asymmetry is deliberate rather than discovered later. If those hosts can set headers, the same two lines belong there.

vercel.json
metalisk, line 14: `report-uri /api/csp-report` is new here, which makes the endpoint newly load-bearing. The handler itself is pre-existing on `dev`, so this is not a defect in this PR — but three things are worth a live check now that reports are actually being sent, ideally by triggering one deliberate violation on a preview deploy:

1. **Do reports arrive at all?** Browsers post CSP reports with `Content-Type: application/csp-report`. Vercel's body parser handles JSON content types; if `request.body` arrives `undefined` for that type, `parseBody` returns `[]` and the handler logs `count: 0` and answers `204`. That failure is completely silent.
2. **The log label is wrong now.** The handler writes `event: 'csp-report-only'`, but this policy is enforcing (`Content-Security-Policy`, not `-Report-Only`). Anyone reading those logs will conclude nothing was blocked.
3. **`isAllowedHost` uses `host.endsWith('adamant-team.vercel.app')`** without a dot boundary, so `notadamant-team.vercel.app` also passes. Low impact — it only gates which host may write log lines — but `=== 'adamant-team.vercel.app' || endsWith('.adamant-team.vercel.app')` is the correct form.

Also note `report-uri` is deprecated in favour of `report-to` / `Reporting-Endpoints`; browsers still honour it, so this is fine for now, just worth knowing it is on a deprecation path.

metalisk (commented):
Full review of the branch, checked out locally on Node 24.19.0 / npm 12.0.2 as declared in `.nvmrc` and `engines`.

Nothing blocking. Six inline notes below — one behavioural change worth a test, one documentation ask on the CSP trade-off, and a few boundaries worth writing down next to the code that has them. Summary follows in a separate comment.

What I ran locally: `typecheck` (vue-tsc + `tsconfig.node.json`), `lint --max-warnings=0`, the full unit suite (**133 files, 823 tests, 0 skipped** — the live network specs ran too), and PWA / Android / Electron builds. All green. I also verified the CSP build gate fires, and the bundle contains no `new Function(`.

metalisk:
## Review summary

Reviewed the whole branch against `dev` — 7 commits, 180 files. I had reviewed the first five commits incrementally as they landed; this pass covers the two CSP/test commits and re-reads the rest as a unit.

**Nothing blocking.** Six inline notes, none of which need to hold the PR.

### What is good

**The CSP work is the strongest part.** `script-src` loses `'unsafe-inline'`, `'unsafe-eval'` and the wildcard host allowances, leaving `'self' 'wasm-unsafe-eval'`. That was only possible because all three dynamic-`Function` sources were actually removed rather than excused: engine.io's `Function('return this')()` is rewritten to `globalThis` by a transform, `setImmediate` got a local polyfill, and `lodash-es` is out of the runtime modules with a contract test holding it there. The build gate is the part I like most — I confirmed it fails the build on a real violation, so this cannot quietly regress.

**Deterministic vectors in the right places.** The BIP39 seed vector is the canonical `abandon … about` one, which is external truth rather than a value captured from the new code — that is what makes the `bip39` → `@scure/bip39` swap trustworthy. Same for the ADM `hash → publicKey → address` chain, the ETH address at the application HD path, the BTC/DOGE/DASH addresses and the identicon colours. The BTC derivation is additionally validated end-to-end: the live specs query real indexers with those addresses.

**The cryptographic boundary is now strict rather than convenient.** `toNaclBytes` rejects anything that is not a `Uint8Array` instead of coercing it. I checked the guard against real values: `Buffer` passes (`[object Uint8Array]`), cross-realm `Uint8Array` still passes — which was the helper's original purpose — while `Int8Array` and `DataView` are rejected. And `fatal: true` UTF-8 is applied exactly where the result feeds `JSON.parse` (KVS values, IDB records) while chat stays tolerant, so a peer sending malformed bytes cannot break rendering.

**Contract tests resolve paths from `import.meta.url`.** That closes a class of bug this branch already hit once, where a cwd-relative path made an entire security contract silently stop running.

**One fix that is invisible in the commit messages.** Removing `detect-browser` also fixed a real bug: `display-mobile.ts` compared the library's OS string against `'android'`, but it returns `'Android OS'` — lowercased, `'android os'` — so the comparison never matched. Android was only ever detected by viewport width, and the two mobile detectors had disagreed with each other for a long time. The follow-up that keeps Android tablets on the mobile layout is the right call and is now commented as a deliberate contract.

**Unit tests now run in CI.** Previously the only workflow with tests was Playwright, which meant a failing unit test was invisible. That gap is closed.

### What to watch

1. **`cloneState` changes a failure mode, not just an implementation** — `structuredClone` throws on proxies and functions and drops prototypes, where `cloneDeep` degraded quietly. Harmless with today's plain `state.chat`, but the throw lands in `PasswordSetDialog`'s fail-safe, which wipes the cache and disables "Stay logged in" without surfacing a cause. A regression test with a non-cloneable value would pin the intended behaviour.
2. **The CSP moves in two directions.** `script-src` is much tighter; `connect-src`, `img-src` and `media-src` widen from a domain allowlist to any `http:`/`https:`/`ws:`/`wss:` origin. The widening is justified by user-configured nodes and is far less dangerous now that inline and `eval` execution are gone — but it belongs in the PR description so nobody re-narrows it later and breaks custom nodes.
3. **`report-uri` is newly wired to a pre-existing handler.** Worth one deliberate violation on a preview deploy to confirm reports actually arrive: browsers send `Content-Type: application/csp-report`, and if the body does not parse, the handler logs `count: 0` and returns `204` silently. The handler also labels its output `csp-report-only` while this policy is enforcing.
4. **`frame-ancestors` cannot travel in a `<meta>` tag.** Vercel sets it as a header, so production is covered; GitHub Pages, deweb and the Tor build get only the meta policy and therefore no framing protection from CSP.
5. **The build gate's boundaries.** It scans emitted chunks, so tree-shaken code is never examined — my first probe passed for exactly that reason — and detection is lexical, so an aliased or computed `Function` reference is not caught. It is a strong guard against accidental reintroduction and should be described as that rather than as proof.
6. **Deprecation note.** `report-uri` is superseded by `report-to` / `Reporting-Endpoints`. Still honoured everywhere, so no action needed now.

### Verification

Checked out locally on **Node 24.19.0 / npm 12.0.2**, matching `.nvmrc` and `engines`:

| check | result |
| --- | --- |
| `typecheck` (vue-tsc + `tsconfig.node.json`) | clean |
| `lint --max-warnings=0` | clean |
| unit suite | **133 files, 823 tests, 0 skipped** |
| PWA / Android / Electron builds | all pass |
| `new Function(` in the PWA bundle | 0 |
| CSP build gate, negative control | fails the build as intended |

The live network specs ran and passed this time, including the BTC indexer path that had been returning 404 for several days — so the suite is genuinely green rather than green-by-skip.

### Earlier rounds

Everything raised in the previous reviews is closed: the polling and single-fetch key-binding paths, the IPFS download bound and its browser enforcement, the KDF worker timeout and fallback, the session E2E with a real reload, the theme sync overwriting an explicit user choice, and the BIP39 contract's magic number. Two were closed more thoroughly than proposed, which is the better outcome.

One standing item that is not this PR's to fix: `live-send-builders.env.spec.js` fails all 26 of its tests whenever a single chain's indexer is unavailable, because the setup for every coin lives in one `beforeAll`, and vitest reports the result as "skipped" — which reads identically to "not configured". Splitting that per coin would make the live suite honest about what actually ran. Happy to open it separately.

.github/workflows/eslint.yml
al-onyxprotocol: Added `npm run build` to the mandatory Quality job after unit tests. This now exercises CSP meta injection, emitted-chunk scanning, service-worker generation, and the strict CSP bundle gate on every PR and protected-branch push.

src/store/plugins/sessionStorage.js
al-onyxprotocol: Updated the PR description to state both reload paths explicitly. Passphrase-only sessions return to full passphrase login after a full reload, while password-protected sessions show the password unlock form.

vite-config/plugins/cspHardeningPlugin.ts
al-onyxprotocol: Documented this residual risk in both the README and PR description. The exception remains limited to WebAssembly compilation for secp256k1; JavaScript string evaluation stays blocked.

vite-config/plugins/cspHardeningPlugin.ts
al-onyxprotocol: Documented the intentional widening and its decentralization rationale in the README and PR description. Custom and self-hosted nodes require arbitrary HTTP(S) and WebSocket origins, while script execution is now substantially narrower.

vite-config/plugins/cspHardeningPlugin.ts
al-onyxprotocol: Clarified this directly beside the scanner and in the README. The gate is defense in depth for reachable emitted chunks and direct lexical calls; runtime CSP remains the enforcement boundary for aliases and computed access.

vercel.json
al-onyxprotocol: Documented the meta/header asymmetry in the plugin, README, and PR description. `frame-ancestors` remains a response-header responsibility; Vercel and Electron enforce it now, while the remaining deployed-host matrix stays tracked in #927.

package.json
al-onyxprotocol: Updated Quick Start to require Node.js `>=24.15.0`, npm `>=12`, and `corepack enable npm`, matching `engines`, `.nvmrc`, and CI.

src/lib/cloneState.ts
al-onyxprotocol: Added regression coverage that supplies a non-cloneable function and asserts that `cloneState` throws instead of degrading silently. The intended fail-safe behavior is now pinned.

src/lib/idb/state.js
al-onyxprotocol: Wrapped the chat clone with a module-specific error log and made `saveState` async so clone failures are exposed as promise rejections to the existing fail-safe. The new test also verifies that no IndexedDB write or readiness commit occurs after a clone failure.

vite-config/plugins/cspHardeningPlugin.ts
al-onyxprotocol: Added the boundary comment beside `findUnsafeRuntimeSources` and expanded the README wording. It now explicitly describes emitted reachable code plus direct lexical calls, without presenting the gate as proof of absence.

vite-config/plugins/cspHardeningPlugin.ts
al-onyxprotocol: Expanded the PR description and README with both directions of the policy change and the reason for the broader scheme sources. This should prevent a future fixed-domain narrowing from breaking user-configured nodes.

vite-config/plugins/cspHardeningPlugin.ts
al-onyxprotocol: Added an explicit comment at the meta policy and documented the host responsibility in the README and PR description. No `frame-ancestors` directive was added to the meta policy because browsers ignore it there.

vercel.json
al-onyxprotocol: Verified this live on the Vercel preview: Chromium blocked a deliberate external script, sent `Content-Type: application/csp-report` with `disposition: enforce`, and received `204` from the endpoint. I also changed the event to `csp-violation`, tightened the host boundary while preserving real Vercel preview hostnames, added bounded raw-stream parsing, rejected empty reports instead of logging count zero, and covered these paths with Node tests. `report-uri` stays for current browser compatibility; migration to Reporting API headers can follow separately.

DEPENDENCY_SECURITY.md
al-onyxprotocol: The exception remains time-bounded to 2026-11-10 in `DEPENDENCY_SECURITY.md` and is also tracked in the full current dependency status on #929. It will be re-evaluated against the Capacitor toolchain before that expiry.

al-onyxprotocol:
## Review follow-up

### Implemented

- Added the production PWA build to the mandatory Quality workflow so CSP injection, emitted chunks, service-worker output, and the strict bundle gate run on every PR
- Made non-cloneable IndexedDB state failures observable by logging the affected module and returning a rejected `saveState` promise; added regression coverage that also proves no partial persistence occurs
- Hardened `/api/csp-report` for raw `application/csp-report` streams, bounded payloads, exact Vercel host boundaries, non-empty reports, and enforcing-policy event labels
- Updated Quick Start to Node.js `>=24.15.0`, npm `>=12`, and Corepack activation
- Documented both credential flows after reload, the deliberately broad custom-node schemes, the `wasm-unsafe-eval` residual, build-gate boundaries, and the meta/header `frame-ancestors` split

### Deliberate decisions

- Kept `wasm-unsafe-eval` because bundled secp256k1 requires WebAssembly compilation; JavaScript string evaluation remains blocked
- Kept arbitrary HTTP(S) and WebSocket origins for user-configured and self-hosted nodes
- Kept `report-uri` for current browser compatibility; Reporting API headers can be migrated separately
- Kept `frame-ancestors` out of the meta policy because browsers ignore it there; Vercel and Electron enforce it through response headers, and #927 tracks the remaining deployed-host matrix

### Verification

- `npm run lint`
- `npm run typecheck`
- Unit suite: 135 files, 829 tests
- `npm run build`, including the CSP generated-bundle gate
- Live Vercel preview violation: Chromium blocked an external script, sent `application/csp-report` with `disposition: enforce`, and received `204`

One clarification from the review summary: `live-send-builders.env.spec.js` is already split into per-coin setup blocks in this branch. The latest live no-broadcast run completed all 26 tests, so one unavailable indexer no longer suppresses unrelated coin coverage.

adamantmm (approved):
lgtm after review

adamant-al (commented):
## Hinweis zum Preview-Comment-Upsert

Kurzer Follow-up nur zu den Surge-Preview-Kommentaren: beim Upgrade auf `thollander/actions-comment-pull-request@v3` muss der Input-Name angepasst werden. Details an den beiden Workflow-Stellen unten.

.github/workflows/preview.yml
```diff
@@ -47,7 +45,7 @@ jobs:
           surge ./dist ${{ env.PREVIEW_DOMAIN }} --token ${{ secrets.SURGE_TOKEN }}
 
       - name: Comment PR 💬
-        uses: thollander/actions-comment-pull-request@v2
+        uses: thollander/actions-comment-pull-request@v3
```
adamant-al, line 48: **Medium — `comment_tag` vs. `comment-tag` (v3)**

Hier wurde auf `thollander/actions-comment-pull-request@v3` angehoben. In **v2** hieß der Input `comment_tag`; in **v3** ist er offiziell `comment-tag` (Bindestrich).

Die folgende Zeile im Workflow bleibt bei:

```yaml
comment_tag: deploy-message
```

Damit greift der Upsert-Tag nicht — jeder Preview-Deploy erzeugt einen **neuen** PR-Kommentar statt den bestehenden zu aktualisieren. Auf diesem PR liegen bereits mehrere identische „Deployed to …surge.sh“-Kommentare von `github-actions[bot]`.

**Fix:**

```yaml
comment-tag: deploy-message
```

Gleiches Muster in `preview-teardown.yml` (`teardown-message`).

.github/workflows/preview-teardown.yml
```diff
@@ -17,13 +17,19 @@ jobs:
...
+        uses: ./.github/actions/setup-node
+
       - name: Tear down ${{ env.PREVIEW_DOMAIN }} preview domain 🥲
         run: |
           npm install --global surge
           surge teardown ${{ env.PREVIEW_DOMAIN }} --token ${{ secrets.SURGE_TOKEN }}
 
       - name: Comment PR 💬
-        uses: thollander/actions-comment-pull-request@v2
+        uses: thollander/actions-comment-pull-request@v3
```
adamant-al, line 32: **Medium — gleiches v3-Input-Problem**

Auch hier: `@v3`, aber darunter weiterhin `comment_tag: teardown-message`.

Bitte umbenennen in:

```yaml
comment-tag: teardown-message
```

Sonst kein stabiles Upsert beim Teardown-Kommentar (analog zu `preview.yml` / `deploy-message`).

api/csp-report.js
```diff
@@ -54,7 +63,43 @@ function parseBody(body) {
   return []
 }
 
-export default function handler(request, response) {
+async function readRequestBody(request) {
+  if (request.body !== undefined) {
```
metalisk, line 67: This branch closes the parsing gap I raised, and the raw-stream path below is the right shape. One thing it cannot settle from the code alone: **whether Vercel ever hands this handler the raw stream** for a CSP report.

Browsers post violation reports with `Content-Type: application/csp-report` (and `application/reports+json` for the Reporting API). Vercel's body parser handles the content types it recognises and leaves `request.body` alone otherwise — but "leaves alone" can mean either `undefined` (stream still readable, this code works) or an already-consumed `{}`.

If it is the second, this early return takes `{}`, `parseBody` yields `[{}]`, `normalizeReportEntry` finds every field `null` and returns `null`, and the handler answers `400` — so reports would be rejected rather than silently logged as `count: 0`. Better than before, but still nothing in the logs.

Worth one deliberate violation against a preview deploy to settle it — e.g. loading an inline `<script>` or an off-origin script on the preview URL and checking the function logs for an `event: "csp-violation"` line. If `{}` turns out to be what arrives, the fix is small: treat an empty-object body the same as `undefined` and fall through to `readRequestBody`.

Unit tests cannot answer this one because the platform's parser is the variable, which is why I am flagging it as a live check rather than a code change.

api/csp-report.js
```diff
@@ -1,8 +1,15 @@
...
-function isAllowedHost(host) {
-  return host === 'dev.adamant.im' || host.endsWith('adamant-team.vercel.app')
+export function isAllowedHost(host) {
+  const normalizedHost = String(host).toLowerCase()
+
+  return (
+    normalizedHost === 'dev.adamant.im' ||
+    normalizedHost === 'adamant-team.vercel.app' ||
+    normalizedHost.endsWith('.adamant-team.vercel.app') ||
+    normalizedHost.endsWith('-adamant-team.vercel.app')
```
metalisk, line 11: The boundary fix is right, and this line is necessary — Vercel preview URLs are `<project>-<team>.vercel.app`, so the deployment host genuinely ends with `-adamant-team.vercel.app` and a dot-boundary check alone would reject every preview.

Recording the one residual case for completeness, not asking for a change: a Vercel team whose slug itself ends in `-adamant-team` would produce hosts matching this suffix too. The impact is limited to who may write a line to the function log — the handler stores nothing, reflects nothing back, and caps both payload size and violation count — and reaching the function at all requires the Host to be a domain Vercel routes to this project, so the platform rejects a spoofed header before this code runs.

If you ever want it exact, the deployment host set is enumerable (`dev.adamant.im`, `adamant-team.vercel.app`, `*.adamant-team.vercel.app`, and `<project>-adamant-team.vercel.app` for the known project names) — but I would leave it as is. Noting it so a future reader does not mistake the loose suffix for an oversight.

metalisk (commented):
Re-reviewed after the two follow-up commits. **All six notes from my previous review are closed**, and three of them more thoroughly than asked:

- `cloneModuleState` logs the failing module and rethrows, and `saveState` is now `async`, so a clone failure always surfaces as a rejection rather than a synchronous throw that only happened to be caught because the single caller sat inside a `.then()`.
- `clearDb` became one atomic transaction across all three stores and now propagates failures instead of swallowing them. I checked all five call sites — `indexedDb.js`, `PasswordSetDialog`, `LoginPasswordForm` and both paths in `Options.vue` — every one handles the rejection, so no unhandled rejection was introduced.
- The CSP gate now runs in CI via `npm run build`, which matters more than it looks: before, the gate only fired for whoever built locally.
- `corepack enable npm` in the composite action *asserts* the resulting version against `packageManager` instead of merely enabling it. That is stricter than what I suggested.
- The E2E race was real — booting the app on `/` let the persistence plugin overwrite the localStorage fixture while the IndexedDB fixture was still being written. Establishing the origin via `/favicon.ico` makes the test exercise what it claims.

The plugin diff is comment-only, so the negative control I ran against the CSP gate previously still holds.

Local verification on Node 24.19.0 / npm 12.0.2: `typecheck` and `lint --max-warnings=0` clean, **135 files / 829 tests, 0 skipped**, PWA / Android / Electron builds pass, zero `new Function(` in the bundle.

Two items remain, both on `api/csp-report.js`, and neither blocks. One is an empirical question about the hosting platform that unit tests cannot answer; the other is a residual I am recording rather than asking you to change. Details inline.

.github/workflows/preview.yml
al-onyxprotocol: Changed the v3 input to `comment-tag: deploy-message`. This restores stable upsert behavior instead of adding a new deployment comment on every synchronization.

.github/workflows/preview-teardown.yml
al-onyxprotocol: Changed the v3 input to `comment-tag: teardown-message`, matching the v3 action contract and restoring stable upsert behavior for teardown comments.

api/csp-report.js
al-onyxprotocol: Confirmed the platform behavior against the current Vercel preview. Chromium delivered a deliberate violation as `application/csp-report` and received `204`; I then repeated the raw content type and payload against the current deployment with a unique blocked URI and again received `204` with an empty response body. Every read, parse, and normalization failure in this handler returns `400`, while `204` is reached only after `process.stdout.write()` emits the `event: "csp-violation"` payload. Therefore Vercel is not handing the handler an already-consumed `{}` and the raw-stream path works as intended, so I did not add the empty-object fallback.

api/csp-report.js
al-onyxprotocol: Agreed. I am leaving the suffix rule unchanged and treating the residual team-slug collision as a documented, bounded trade-off: Vercel routing constrains reachability, the endpoint stores and reflects nothing, and both payload size and logged violation count are capped.

adamant-al (approved):
huge work and it's done ✅

metalisk (approved):

Issue #962, opened 2026-08-09
[Task] Bounded request timeouts and observed-latency node ranking, without weakening decentralization


adamantmm:
### Summary

Give ADM node and service requests a bounded, connection-aware timeout, and rank nodes by observed latency of real requests instead of by ping — without weakening decentralization when "Choose the fastest node" is off.

#### Problem 1 — ADM requests have no timeout at all

`AdmNode.buildClient()` returns `axios.create({ baseURL: this.url })`. Axios defaults to `timeout: 0`, meaning wait forever. `RateInfoService` has the same gap. The coin clients do not:

| Client | Request timeout |
|---|---|
| `AdmNode` | **none** |
| `RateInfoService` | **none** |
| `EthIndexer` | `getConnectionAwareTimeout(...)` |
| `createBtcLikeClient` (BTC, DOGE, DASH) | `getConnectionAwareTimeout(...)` |
| `IpfsNode` | 10 minutes, appropriate for file transfers |

So the client used for every chat, account and transaction request is the one with no ceiling.

This also disables failover. `Client.requestWithRetry()` moves to another node only when a request **throws**. With no timeout it never throws — it waits. A node that responds slowly, or not at all while keeping the connection open, stalls the application indefinitely instead of being replaced.

Observed: a single `/api/chatrooms/{address}?limit=25` took 54 seconds against a v0.8.3 node during login. The same request measured 3.0–3.6 s against that node minutes later and 0.6–0.7 s against a v0.10.3 node — so the node was having a bad moment, and nothing was going to cut it off.

Node version matters too, for the same account with 283 chats:

| Node | Version | `/api/chatrooms?limit=25` |
|---|---|---|
| tegmine.adm.im | 0.8.3 | 3635 / 2972 / 2991 ms |
| clown.bbry.org | 0.9.0 | 7576 ms |
| bid.adamant.im | 0.10.3 | 1763 ms |
| endless.adamant.im | 0.10.3 | 727 / 597 / 620 ms |

Roughly a 5× improvement from 0.8.3 to 0.10.3. That is the node operators' side; the client side is that it should not have waited 54 seconds for any of them.

#### Why the timeout is not a one-line change

A short timeout is worse than none for users on poor connections — every request would fail instead of eventually succeeding.

- the base value has to be generous, on the order of 20–30 s, not the 10 s used for health checks
[...]

PR #966, opened 2026-08-10
Fix: Complete security hardening part 3

### LINKED ISSUE #958: [Composite, Task] Full security audit and coordinated hardening release

adamant-al:
## Description

This is the third security-hardening batch from #958, based on the full audit in the branch-local
`.ai-tasks` documents. It completes the next five code-level audit decisions for 4.12, hardens
Android storage boundaries, makes protocol-message filtering consistent, and fixes the remaining
public encryption-helper boundary. It also stops client-only virtual chats from polling impossible
node history.

### Android data and file-sharing boundaries

- Disables Android cloud backup and device-to-device transfer for app files, databases, shared preferences and external app data
- Supplies both legacy `fullBackupContent` rules and Android 12+ `dataExtractionRules` instead of relying on `allowBackup` alone
- Narrows the generic Capacitor capture provider to its app-specific `Pictures/` destination
- Overrides FileOpener's broad defaults with private `files/attachments/` only
- Reduces untrusted attachment names to a basename before native writes, preventing `../` traversal outside the dedicated directory
- Adds contract tests for the manifest, both backup-rule formats and both FileProvider path files

Ordinary attachment and QR-image selection still uses Android's system picker and is not restricted
by these FileProvider paths. Native attachment downloads remain private to ADAMANT until FileOpener
grants an external viewer temporary read access; no public Downloads or media-storage permission is
added.

### AIP-6 signal-message visibility

- Adds one `isChatTransactionVisible` rule for socket delivery, all three REST chat paths, Vuex ingress and direct quoted-message fetches
- Filters protocol signal messages before public-key handling, decryption, normalization and bot-command processing
- Does not seed the verified-key cache from hidden signal transactions; a later visible transaction performs the normal binding or on-demand lookup
- Advances REST pagination by the raw number of processed transactions, so signal-only pages neither loop nor truncate later visible history
- Advances the initial account-wide polling watermark only through successfully processed visible room transactions, so a trailing signal cannot hide an earlier message
- Advances the initial chat-list offset independently by the raw fetched room count, including when the first page contains only hidden signals
- Keeps signal processing local and adds no node request or network round trip

### Cryptographic and binary attachment boundaries

- Normalizes documented hex or byte-array inputs at the encode, decode and address-derivation boundaries
- Accepts documented hexadecimal and `Uint8Array` inputs while rejecting malformed or odd-length hex instead of silently coercing it
- Hardens the unused legacy attachment-upload action with an explicit `application/octet-stream` Blob; the released chat-send path already used binary Blobs
- Adds deterministic round-trip, malformed-input and exact binary `FormData` coverage

ADM-SEC-021 is closed as not a defect. AIP-3 intentionally frames the authenticated KVS JSON with
brace-free random padding, so first/last-brace parsing remains the interoperable multi-client wire
format. A regression test covers payload values that contain nested braces; no protocol change is
introduced.

### Repository security settings

ADM-SRV-003 was completed alongside this branch through repository administration:

- Vulnerability alerts and Dependabot security updates are enabled
- Secret scanning and repository push protection are enabled
- CodeQL continues to publish analyses from the default `dev` branch
- Dependabot version updates and the mandatory security workflow target `dev`

### Virtual chat polling

The built-in welcome chat is now explicitly marked as client-only. Its history cursor closes locally,
so opening it no longer requests invalid
`/api/chatrooms/<address>/chats.virtual.welcome_message_title` history from a node.

## Related issues

- Part of #958
- Node-side WebSocket origin validation remains tracked in Adamant-im/adamant#203
- ADM request timeouts and failover remain tracked in #962

## How to test

Run with Node 24 and the pinned npm 12 release:

```bash
corepack npm ci
corepack npm run lint
corepack npm run typecheck
corepack npm run test -- --run
corepack npm run build
corepack npm run android:prebuild
```

On Android:

1. Open the welcome chat and confirm no `chats.virtual.*` history request is made
2. Select images and files from the system picker, including Downloads or a document provider
3. Download an attachment and confirm it opens in the selected external viewer
4. Build the signed APK and AAB through the `Capacitor [Android]` workflow

## Verification performed

- Node 24.19.0 and npm 12.0.2
- ESLint: clean with zero warnings
- Typecheck: clean (`vue-tsc` and Node/Vite configs)
- Unit tests: 139 files, 856 tests passed
- PWA production build: passed
- Android frontend prebuild: passed
- Signed native APK/AAB workflow: passed on this branch; `android-release-app` contains both signed artifacts
- Repository secret scanning and push protection: verified enabled through the GitHub API
- `git diff --check`: clean

## Checklist

- [x] Android secrets are excluded from backup and device transfer across supported rule formats
- [x] FileProvider access is limited to the exact app-owned paths used by capture and FileOpener
- [x] System picker access to user-selected files remains unchanged
- [x] Signal-message filtering covers REST, socket, store and quoted-message entry points
- [x] Hidden protocol messages cannot stall pagination or add requests
- [x] Hex and byte-array encryption inputs are covered by interoperability tests
- [x] No passphrase, private key, plaintext attachment or token logging introduced
- [x] No analytics, tracking or centralized endpoint introduced
- [x] Signed native APK and AAB build verified for this branch


android/app/src/main/AndroidManifest.xml
```diff
@@ -1,7 +1,9 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android">
     <application
-        android:allowBackup="true"
+        android:allowBackup="false"
+        android:dataExtractionRules="@xml/data_extraction_rules"
+        android:fullBackupContent="@xml/backup_rules"
```
dev-adamant-im, line 6: **Dobrze — backup wyłączony na obu warstwach**

`allowBackup="false"` plus jawne `fullBackupContent` i `dataExtractionRules`. Na nowszych API same `allowBackup` nie wystarcza przeciw device-to-device transfer — tu oba formaty reguł są spójnie wykluczające. Solidne zamknięcie ekspozycji lokalnych sekretów przez backup.

android/app/src/main/res/xml/backup_rules.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<full-backup-content>
+    <exclude domain="root" path="." />
```
dev-adamant-im, line 3: **Dobrze — legacy full-backup exclude**

Wykluczenie wszystkich domen (`root` / `file` / `database` / `sharedpref` / `external`) bez `<include>`. Razem z `data_extraction_rules.xml` nie zostawia luk między API levels.

android/app/src/main/res/xml/data_extraction_rules.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<data-extraction-rules>
+    <cloud-backup>
```
dev-adamant-im, line 3: **Dobrze — cloud-backup i device-transfer**

Oba bloki z pełnym exclude. To właściwa odpowiedź na Android 12+ data extraction, nie tylko na stary cloud backup.

android/app/src/main/res/xml/file_paths.xml
```diff
@@ -1,5 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
 <paths xmlns:android="http://schemas.android.com/apk/res/android">
-    <external-path name="my_images" path="." />
-    <cache-path name="my_cache_images" path="." />
-</paths>
\ No newline at end of file
+    <external-files-path name="captured_images" path="Pictures/" />
```
dev-adamant-im, line 3: **Dobrze — wąski Capacitor FileProvider**

Zamiast `external-path` / `cache-path` z `path="."` zostaje tylko app-specific `external-files-path` → `Pictures/` pod capture. Duże zmniejszenie powierzchni grantów URI.

android/app/src/main/res/xml/file_opener_paths.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<paths xmlns:android="http://schemas.android.com/apk/res/android">
+    <files-path name="downloaded_attachments" path="attachments/" />
```
dev-adamant-im, line 3: **Wyjaśnienie — to NIE jest brak wiring w app manifest (false positive Bugbot)**

Ten plik **nie musi** być podpięty w `AndroidManifest.xml` aplikacji.

`@capacitor-community/file-opener` deklaruje własny provider (`${applicationId}.file.opener.provider`) z `android:resource="@xml/file_opener_paths"`. Zasób aplikacji o tej samej nazwie **nadpisuje** domyślny XML pluginu (tam było szerokie `path="."` na files/cache/external).

Tu zostaje wyłącznie `files-path` → `attachments/`, zgodnie z `Directory.Data` + `getNativeAttachmentPath`. Integracja jest poprawna (fail-closed względem starego defaultu pluginu).

src/lib/androidSecurityContract.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+
+      expect(legacyRules).toContain(exclusion)
+      expect(extractionRules.split(exclusion)).toHaveLength(3)
+    }
+
+    expect(legacyRules).not.toContain('<include')
+    expect(extractionRules).not.toContain('<include')
+  })
+
+  it('limits both FileProviders to their application-specific directories', async () => {
```
dev-adamant-im, line 37: **Dobrze — kontrakt na oba FileProvidery**

Test pinuje wąskie ścieżki i zakazuje `external-path` / `cache-path` / `path="."`. Opcjonalnie później: asercja, że app `file_opener_paths.xml` jest ściślejszy niż default z `node_modules/@capacitor-community/file-opener` — żeby kolejny review nie uznał override za „martwy” plik.

src/lib/files/nativeAttachmentPath.ts
```diff
@@ -0,0 +1,10 @@
+export const NATIVE_ATTACHMENT_DIRECTORY = 'attachments'
+
+const getBaseName = (filename: string) => filename.replaceAll('\\', '/').split('/').at(-1) ?? ''
+
+export function getNativeAttachmentPath(filename: string) {
+  const baseName = getBaseName(filename)
+  const safeName = baseName && baseName !== '.' && baseName !== '..' ? baseName : 'unnamed'
+
+  return `${NATIVE_ATTACHMENT_DIRECTORY}/${safeName}`
+}
```
dev-adamant-im, line 10: **Dobrze — sanitizacja ścieżki przed native write**

Basename po normalizacji separatorów + odrzucenie `.` / `..`. Nazwa z `FileAsset` (kontrolowana przez nadawcę) nie wyjdzie poza `attachments/`. Kolizje nazw to UX/availability, nie bypass sandboxu.

src/lib/chat/helpers/isChatTransactionVisible.js
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * Returns whether a raw transaction is allowed to enter a user-visible chat path.
+ * AIP-6 signal messages are protocol control messages and must stay hidden in dialogs.
+ *
+ * @param {object} transaction Raw ADM transaction
+ * @returns {boolean}
+ */
+export function isChatTransactionVisible(transaction) {
+  return transaction?.asset?.chat?.type !== MessageType.SIGNAL_MESSAGE
```
dev-adamant-im: **Note — ścisłe porównanie typu**

`!== MessageType.SIGNAL_MESSAGE` (`3`) przepuści `type: "3"` (string). Ten sam wzorzec był wcześniej w sockecie.

Dla złośliwego węzła to nie jest silniejsze niż podanie `type: 1`; key-binding i tak chroni podmianę klucza. Poniżej medium.

**Defense-in-depth (opcjonalnie):**

```js
Number(transaction?.asset?.chat?.type) === MessageType.SIGNAL_MESSAGE
```

z guardem na `NaN`, albo odwrotnie: widoczne tylko gdy `Number(...) !== 3` po walidacji.

src/store/plugins/socketsPlugin.js
```diff
@@ -1,11 +1,13 @@
 import socketClient from '@/lib/sockets'
 import { cacheVerifiedPublicKey, decodeChat, getPublicKey } from '@/lib/adamant-api'
 import { isStringEqualCI } from '@/lib/textHelpers'
-import { MessageType } from '@/lib/constants'
 import { logger } from '@/utils/devTools/logger'
+import { isChatTransactionVisible } from '@/lib/chat/helpers/isChatTransactionVisible'
 
 function subscribe(store) {
   socketClient.subscribe('newMessage', (transaction) => {
+    if (!isChatTransactionVisible(transaction)) return
```
dev-adamant-im, line 9: **Dobrze — filtr przed key binding / decode**

Signal odpada zanim `cacheVerifiedPublicKey` / `getPublicKey` / `decodeChat`. To zamyka zbędne requesty i wcześniejszą lukę, gdzie signal mógł wejść w ścieżkę kryptograficzną. Potwierdzone testem socketa.

src/lib/adamant-api/index.js
```diff
@@ -544,8 +545,10 @@ export function getChats(from = 0, offset = 0, orderBy = 'desc') {
   // https://github.com/Adamant-im/adamant/wiki/API-Specification#get-chat-transactions
   return client.get('/api/chats/get/', params).then((response) => {
     const { count, transactions, nodeTimestamp } = response
+    const fetchedCount = transactions.length
+    const lastProcessedHeight = transactions[transactions.length - 1]?.height || 0
 
-    const promises = transactions.map((transaction) => {
+    const promises = transactions.filter(isChatTransactionVisible).map((transaction) => {
```
dev-adamant-im, line 551: **Dobrze — filtr + raw pagination cursor**

`fetchedCount` / `lastProcessedHeight` z **surowej** strony, a nie z listy już odfiltrowanej. Strony samych signal nie zapętlają pollingu ani nie ucinają późniejszej widocznej historii. Właściwy kontrakt względem AIP-6.

src/lib/adamant-api/index.js
```diff
@@ -770,8 +775,14 @@ export async function getChatRooms(address, params) {
     ...defaultParams,
     ...params
   })
+  const fetchedCount = chats.length
```
dev-adamant-im, line 778: **Dobrze — getChatRooms też awansuje po raw count**

`fetchedCount = chats.length` przed filtrem `lastTransaction`. Offset listy pokoi nie staje w miejscu, gdy ostatnia tx to signal.

src/lib/adamant-api/index.js
```diff
@@ -852,6 +862,10 @@ export async function getChatRoomMessages(address1, address2, paramsArg, recursi
     }
 
     const decodedMessages = messages.flatMap((message) => {
+      if (!isChatTransactionVisible(message)) {
```
dev-adamant-im, line 871: **Dobrze — getChatRoomMessages filtruje przed decode**

Signal nie wchodzi w binding klucza / `decodeChat` także na historii pokoju. Spójne z socketem i `getChats`.

src/store/modules/chat/index.js
```diff
@@ -752,7 +759,7 @@ const actions = {
    * @param {Message[]} messages Array of messages
    */
   pushMessages({ commit, rootState, dispatch }, messages) {
-    const normalizedMessages = messages.map(normalizeMessage)
+    const normalizedMessages = messages.filter(isChatTransactionVisible).map(normalizeMessage)
```
dev-adamant-im, line 765: **Dobrze — defense-in-depth na wejściu Vuex**

`pushMessages` / `unshiftMessages` filtrują ponownie. Nawet jeśli jakiś caller pominie filtr wyżej, signal nie wejdzie do UI / bot-commands.

src/components/AChat/QuotedMessage.vue
```diff
@@ -86,6 +90,13 @@ async function fetchTransaction(transactionId: string, address: string) {
     )
   }
 
+  if (!isChatTransactionVisible(rawTx)) {
+    throw new ValidationError(
+      `QuotedMessage: Signal messages are hidden: txId: ${transactionId}`,
+      ErrorCodes.MESSAGE_NOT_FOUND
+    )
+  }
```
dev-adamant-im, line 98: **Dobrze — cytaty po id też ukrywają signal**

Fetch po tx id z węzła przechodzi przez tę samą regułę przed decode.

**Note (residual, poza zakresem blockera):** ścieżki typu `decodeTransaction` / szczegóły tx po bezpośrednim id oraz odczyt z cache bez ponownego checka visibility nadal mogą istnieć poza tym PR. Nie jest to regresja tego batcha — ewentualny follow-up dla pełnej jednolitości.

src/lib/chat/meta/chat-meta.js
```diff
@@ -12,6 +12,7 @@ export const ADAMANT_CHATS_META = {
     displayName: 'chats.virtual.welcome_message_title',
     ableToSendMessage: false,
     staticChat: true,
+    virtual: true,
```
dev-adamant-im, line 15: **Dobrze — welcome jako client-only**

`virtual: true` oznacza brak sensownej historii węzła dla `chats.virtual.welcome_message_title`.

src/store/modules/chat/index.js
```diff
@@ -718,6 +720,11 @@ const actions = {
    * @returns {Promise}
    */
   getChatRoomMessages({ rootState, dispatch, commit, getters }, { contactId, perPage = 25 } = {}) {
+    if (isVirtualChat(contactId)) {
+      commit('setChatOffset', { contactId, offset: -1 })
+      return Promise.resolve()
+    }
```
dev-adamant-im, line 729: **Dobrze — brak pollingu virtual chat**

Offset zamykany lokalnie (`-1`), bez requestu `/api/chatrooms/.../chats.virtual.*`. Mniej szumu w logach węzła i brak wycieku „adresu” client-only chatu do sieci.

src/lib/adamant.js
```diff
@@ -35,6 +35,31 @@ const toNaclBytes = (value) => {
...
+  if (typeof value === 'string') {
+    if (!isHexadecimalString(value)) {
+      throw new TypeError(`${label} must be an even-length hexadecimal string or Uint8Array`)
+    }
+
+    return toNaclBytes(hexToBytes(value))
+  }
+
+  return toNaclBytes(value)
+}
```
dev-adamant-im, line 61: **Dobrze — jawna granica hex / bytes (ADM-SEC-027)**

`Uint8Array.from(hexString)` wcześniej dawało ciche zera zamiast odrzucenia. Teraz: nawet długość + charset hex, inaczej `TypeError`. To właściwa naprawa footguna na publicznej granicy encode.

src/lib/adamant.js
```diff
@@ -368,20 +393,18 @@ adamant.verify = function (hash, signatureBuffer, publicKeyBuffer) {
...
   const nonce = nacl.randomBytes(24)
 
-  if (typeof recipientPublicKey === 'string') {
-    recipientPublicKey = hexToBytes(recipientPublicKey)
-  }
-
   const plainText = typeof msg === 'string' ? textEncoder.encode(msg) : toNaclBytes(msg)
-  const DHPublicKey = ed2curve.convertPublicKey(toNaclBytes(recipientPublicKey))
-  const DHSecretKey = ed2curve.convertSecretKey(toNaclBytes(privateKey))
+  const DHPublicKey = ed2curve.convertPublicKey(
```
dev-adamant-im, line 402: **Dobrze — encodeMessage używa normalizacji**

I publiczny, i prywatny klucz przechodzą przez `normalizeCryptoBytes`. Interop hex (API) + `Uint8Array` (wewnętrznie) zachowany; malformed hex pada głośno. Analogicznie w `encodeBinary`.

dev-adamant-im (commented):
## Podsumowanie review (Security Hardening Part 3)

**Werdykt:** Comment / Approve with notes — w diffie **nie ma** findingsów medium+ z realistyczną exploitowalnością. Bugbot zgłosił high na FileProvider; to **fałszywy alarm** (patrz komentarz przy `file_opener_paths.xml`).

### Co jest dobre
- **Android backup / transfer:** `allowBackup="false"` + legacy `fullBackupContent` + Android 12+ `dataExtractionRules` (cloud i device-transfer)
- **FileProvider:** wąski `Pictures/` dla kamery; override ścieżek FileOpener do samego `attachments/` (zamiast domyślnego `path="."` w pluginie)
- **Traversale nazw plików:** basename + odrzucenie `..` przed zapisem native
- **AIP-6:** jedna reguła `isChatTransactionVisible` na REST, socket (przed key binding/decode), Vuex i cytaty; paginacja po raw `fetchedCount`
- **Crypto:** `normalizeCryptoBytes` na granicy `encodeMessage` / `encodeBinary` — bez cichego coerowania hex → zera
- **Upload załączników:** `Blob` `application/octet-stream` bez konwersji na string
- **Virtual welcome chat:** brak pollingu niemożliwej historii węzła

### Na co zwrócić uwagę
1. Porównanie signal type przez `!==` przepuści string `"3"` (defense-in-depth, poniżej medium)
2. Ścieżki tx-by-id / cache poza tym PR nadal bez tej samej filtracji visibility
3. Checklist: signed APK/AAB jeszcze nieodhaczone — warto domknąć przed releasem Android

### Zestawienie findings

| Severity | Miejsce | Finding |
|----------|---------|---------|
| — | — | **Brak medium / high / critical** |
| Note | `isChatTransactionVisible.js` | Ścisłe `!== 3` nie łapie `type: "3"` |
| Note | poza głównymi ścieżkami czatu | `decodeTransaction` / detale po id — residual |
| Process | Checklist PR | Signed native build jeszcze pending |
| Clarification | `file_opener_paths.xml` | Wiring przez provider pluginu + resource override — OK, nie brak w app manifest |

Komentarze inline poniżej: findings, clarifications i pozytywne uwagi.

src/lib/attachment-api/index.ts
```diff
@@ -44,7 +44,10 @@ export class AttachmentApi {
       hexToBytes(publicKey),
       this.myKeypair.privateKey
     )
-    formData.append('file', binary)
+    formData.append(
+      'file',
+      new Blob([new Uint8Array(binary)], { type: 'application/octet-stream' })
+    )
```
dev-adamant-im, line 50: **Dobrze — binarny upload bez string coercion**

`Blob` + `application/octet-stream` utrzymuje bajty ciphertext (w tym >127). Bez tego FormData/`append` potrafi zepsuć payload. Test round-trip w `downloadBound.spec.ts` pinuje kontrakt.

metalisk (commented):
Reviewed the branch locally against `dev`. Compact and well-scoped stage — nothing blocking.

Four inline notes below: one question about data already on IPFS that should be answered before merge, one asymmetry left over from the previous stage, and two behavioural details that are not in the description. Summary follows in a separate comment.

Local verification on Node 24.19.0 / npm 12.0.2: `typecheck` (vue-tsc + `tsconfig.node.json`) and `lint --max-warnings=0` clean, **139 files / 852 tests, 0 skipped**, PWA / Android / Electron builds all pass.

src/lib/attachment-api/index.ts
```diff
@@ -44,7 +44,10 @@ export class AttachmentApi {
       hexToBytes(publicKey),
       this.myKeypair.privateKey
     )
-    formData.append('file', binary)
+    formData.append(
+      'file',
+      new Blob([new Uint8Array(binary)], { type: 'application/octet-stream' })
```
metalisk, line 49: This is the most consequential change in the PR and it is not mentioned in the commit message. Worth pulling out.

`FormData.append` only keeps binary data when the value is a `Blob` or `File`. For anything else the spec converts it with `String(value)`, so the previous `formData.append('file', binary)` with a `Uint8Array` did not upload bytes — it uploaded their decimal text. Confirmed on native `FormData`:

```js
const fd = new FormData()
fd.append('old', new Uint8Array([0, 127, 128, 255]))
fd.get('old') // → "0,127,128,255"  (a string, not a Blob)
```

The `Blob` wrapper is the right fix, and the new test name (`without string coercion`) says you found exactly this.

The question that should be answered before merge is about the data already on IPFS. `git log -S` puts this line back at `321298e9`, i.e. it has been there since attachments were introduced. If uploads really were sending the ASCII rendering of the ciphertext, then every attachment uploaded by a released build is undecryptable, and that belongs in the release notes rather than in a silent fix.

If instead something downstream was compensating — an axios adapter, a `transformRequest`, the multipart serializer — then the impact is nil and this is just correctness. Either answer is fine; the reason for asking is that the two lead to very different release notes.

Worth checking against a real node once, by uploading an attachment from a build without this patch and inspecting what the IPFS object actually contains.

src/lib/adamant.js
```diff
@@ -35,6 +35,31 @@ const toNaclBytes = (value) => {
...
+    const code = character.charCodeAt(0)
+
+    return (code >= 48 && code <= 57) || (code >= 65 && code <= 70) || (code >= 97 && code <= 102)
+  })
+}
+
+// Public crypto helpers historically accept both the hex representation used by the API and
+// byte arrays used internally. Normalize that documented boundary explicitly: Uint8Array.from()
+// would otherwise turn a hex string into zeroes instead of rejecting it.
+const normalizeCryptoBytes = (value, label) => {
```
metalisk, line 51: This closes the last part of the `string | Uint8Array` contract issue from the previous stage — a malformed hex string now throws instead of becoming a zero-filled key. Good.

It is only wired into the two encode paths, though. The decode side still calls bare `hexToBytes`:

- `decodeMessage` — `msg`, `nonce`, `senderPublicKey`, `privateKey`
- `decodeValue` — `source`, `nonce`
- `decodeBinary` — `nonce`, `senderPublicKey`
- `getAddressFromPublicKey` — `publicKey`

`hexToBytes` uses `parseInt(pair, 16)`, so a non-hex pair yields `NaN`, and `Uint8Array.from([NaN])` stores `0` — the same silent coercion `normalizeCryptoBytes` was written to stop.

The failure modes on that side are safe, which is why I am not calling this a defect: a wrong key makes `box.open` return `null` and the message renders as undecryptable, and a malformed public key produces an address that fails the binding check. Nothing is accepted that should not be.

But the asymmetry will read as an oversight to the next person, and the helper already exists — routing the decode paths through it costs a few lines and makes the boundary uniform in both directions. The one thing to keep in mind is that `decodeMessage` is on the hot path for every rendered message, so the validation cost is per-message; `isHexadecimalString` is O(n) over a 64-character string, which is negligible, but worth a conscious decision rather than an accident.

src/store/modules/chat/index.js
```diff
@@ -699,13 +701,13 @@ const actions = {
 
     return admApi
       .getChatRooms(rootState.address, { offset, limit: perPage })
-      .then(({ messages }) => {
+      .then(({ messages, fetchedCount = perPage }) => {
```
metalisk: The pagination rework is right — counting raw results rather than visible ones is what stops a page made entirely of hidden protocol messages from ending the walk early, and `lastProcessedHeight` stops polling from re-requesting the same hidden messages forever. Both traps avoided.

One note on this default value. `getChatRooms` always returns `fetchedCount`, so `= perPage` never fires today. But the value chosen for a fallback decides what happens if it ever stops being returned — a refactor, a mocked client, an older node shape — and `perPage` means *"assume a full page arrived"*. With zero messages and no `fetchedCount`, `fetchedCount <= 0` is false and the offset keeps advancing instead of terminating with `setOffset(-1)`.

`= 0` is the fail-safe direction: if the field is missing, stop rather than loop. The same reasoning applies to `result.fetchedCount ?? length` in `lib/chat/helpers/getChats.js`, which already defaults to the safe side.

Not a live bug — just the difference between a fallback that degrades to "stop" and one that degrades to "keep going".

src/store/plugins/socketsPlugin.js
metalisk, line 9: Moving the check to the top of the handler is cleaner than the old late gate, and it means signal messages no longer reach `pushMessages` through any path. That part is right.

It does change one thing that is not in the description: the handler used to run `cacheVerifiedPublicKey` on every incoming transaction, including signal messages, so a counterparty's public key could be verified and cached from a signal message alone. With the early return that no longer happens.

The effect is small — the key is fetched on demand later, and skipping it means one less thing accepted from an unverified source, so if anything this is the safer direction. But it is a behavioural change to the key cache, and if anyone measures request counts (the previous stage added tests that assert exact request counts precisely because this path is supposed to cost nothing extra), this is where an extra `getPublicKey` round trip could appear for a partner whose only recent traffic was signal messages.

Worth one line in the description, or a comment here, so it reads as intended rather than incidental.

metalisk:
## Review summary

30 files, one commit, checked out and verified locally. **Nothing blocking.** Four inline notes, one of which is a question rather than a change request.

### What is good

**The FileProvider narrowing is a real fix, not hygiene.** `@capacitor-community/file-opener` ships a default `file_opener_paths.xml` that includes `<external-path name="external" path="." />` — the whole external volume reachable through a content provider. The app now supplies its own copy, and since app resources override library resources of the same name, the narrow one wins. I checked the seam rather than assuming it: the download path writes to `Directory.Data` + `attachments/<name>`, which is `getFilesDir()/attachments/`, and the new rule is `<files-path path="attachments/" />`. They match, and `FileOpener.open` in `AChatImageModal` is the only consumer.

**`allowBackup="false"` plus both rule files is not redundant.** On Android 12+ `allowBackup` does not govern device-to-device transfer; that is what the `device-transfer` section of `data_extraction_rules` is for. Excluding every domain in both places is the correct belt-and-braces, and the contract test pins the manifest attributes, both rule files and the absence of any `<include>`.

**The pagination rework avoids two traps that are easy to walk into.** Filtering messages out of a paged API breaks in two directions at once: counting visible results makes a page of entirely hidden messages look like the end of the list, and taking the height from the last *visible* transaction makes polling re-request the hidden ones forever. `fetchedCount` is computed before the filter and `lastProcessedHeight` comes from the raw response, so both are handled.

**Reactions are not affected.** That was the first thing I checked, because reactions are the obvious candidate for collateral damage when message types start being filtered. The application only ever sends `RICH_CONTENT_MESSAGE` and `BASIC_ENCRYPTED_MESSAGE`; `SIGNAL_MESSAGE` is never produced, only received. Reactions travel as rich content with `reactto_id` / `react_message`, so the filter cannot reach them.

**`normalizeCryptoBytes` finishes the boundary work from the previous stage.** The documented `string | Uint8Array` contract now actually holds on the encode paths, and a malformed hex string throws rather than turning into a zero-filled key.

**`QuotedMessage` handles hidden messages as "not found" rather than rendering them**, reusing the existing `ValidationError` path instead of inventing a new state.

### What to watch

1. **Attachment uploads were sending text, not bytes** — `formData.append('file', binary)` with a `Uint8Array` coerces to `"0,127,128,255"` on native `FormData` (verified). The `Blob` wrapper fixes it. The open question is what happened to attachments already uploaded by released builds, since the line dates back to `321298e9`. That answer decides whether this needs a release note.
2. **Validation is applied to encode but not decode** — `decodeMessage`, `decodeValue`, `decodeBinary` and `getAddressFromPublicKey` still use bare `hexToBytes`, which silently turns bad hex into zero bytes. Both fail safely, so this is about making the boundary uniform rather than fixing a hole.
3. **`fetchedCount = perPage` is a fallback that degrades toward "keep paginating"** rather than "stop". Unreachable today; the concern is which way it fails if the field ever goes missing.
4. **The socket handler no longer caches public keys from signal messages**, because the visibility check moved ahead of `cacheVerifiedPublicKey`. Safer direction, but it is a change to the key cache that is not in the description.

Deliberately not raised: `file_paths.xml`. Narrowing it is safe and it stays for future use.

### Verification

Node 24.19.0 / npm 12.0.2, matching `.nvmrc` and `engines`:

| check | result |
| --- | --- |
| `typecheck` (vue-tsc + `tsconfig.node.json`) | clean |
| `lint --max-warnings=0` | clean |
| unit suite | **139 files, 852 tests, 0 skipped** |
| PWA / Android / Electron builds | all pass |

CI is green apart from the Playwright job, which was still running at the time of review.

src/lib/attachment-api/index.ts
adamant-al: Confirmed: there is no released-data impact. In `v4.9.0` and every later release, the active chat attachment sender uses `src/lib/files/upload.ts`, which already wraps ciphertext in a `Blob` and posts the `files` field expected by the IPFS service. `attachment/uploadAttachment` has no callers, so this only hardens a dormant legacy action. Existing IPFS objects need no migration and no compatibility release note; I clarified this distinction in the PR description.

src/lib/adamant.js
adamant-al: Fixed: every documented hex-capable decode and address-derivation boundary now routes through `normalizeCryptoBytes`: `decodeMessage`, `decodeValue`, `decodeBinary`, and `getAddressFromPublicKey`. Malformed hex now throws before NaCl or hashing, while valid hex and byte inputs keep their existing round trips. Added regression coverage for each affected API.

src/store/plugins/socketsPlugin.js
adamant-al: Documented as intentional in the PR description: hidden signal transactions do not seed the verified-key cache. A later visible transaction performs the normal local binding or on-demand lookup. This avoids treating a hidden protocol message as a cache source and keeps processing of the signal itself at zero requests.

src/components/AChat/QuotedMessage.vue
adamant-al: I am keeping `decodeTransaction` and the generic ADM transaction-details route unchanged. They are explicit transaction-inspection APIs reached by ID, not chat ingress, and changing their return contract to hide a transaction would affect the transaction-details surface beyond ADM-SEC-026. The chat paths that can surface a quoted message are filtered before decode; explorer-level suppression would be a separate product/protocol decision.

src/store/plugins/socketsPlugin.js
metalisk: Re-checked after `0387eef4`: this one is unchanged, which is fine as code — I am flagging it again only because it is still absent from the description.

To restate it precisely, since the surrounding diff has moved: before this PR the visibility check sat at the very end of the handler, so a signal message still went through `cacheVerifiedPublicKey` and could populate the public-key cache for that counterparty. With the check at the top, it no longer does.

That is the safer direction — one less thing accepted from an unverified source — so I am not asking for a code change. What is worth a line somewhere is that the key cache is now fed by fewer message types than before, because the previous stage deliberately added tests asserting exact request counts on these paths. If a partner's only recent traffic is signal messages, their key now resolves through a `getPublicKey` round trip that previously would not have happened. Nobody looking at those request-count assertions later will guess that from the diff.

Either a sentence in the PR description or a comment above the guard would do.

metalisk:
## Re-review after `0387eef4`

Two of my four notes are closed, and both correctly.

**Validation now covers the decode side** — `normalizeCryptoBytes` is wired into `decodeMessage`, `decodeValue`, `decodeBinary` and `getAddressFromPublicKey`, so the encode/decode asymmetry is gone.

I checked the consequence of that change rather than just the change itself, because it converts silent coercion into a throw on the path every rendered message takes. All four consumers contain it:

| path | containment |
| --- | --- |
| `getChats` (polling) | per-transaction `.catch`, message dropped |
| `getChatRoomMessages` | `try/catch` inside the `flatMap` |
| `socketsPlugin` | `.catch` at the end of the chain |
| `QuotedMessage` | converted to `ValidationError` |

So one malformed field from a node costs one message, not the page. Worth stating explicitly because without those wrappers this change would have turned a single bad hex string into a failed chat list.

The boundary is also drawn deliberately rather than blanket-applied: hex-or-bytes where the contract documents both, bytes-only elsewhere — and `rejects string input at byte-only cryptographic boundaries` pins that distinction so it does not erode.

**`fetchedCount = 0`** now degrades toward stopping instead of paginating. Closed.

---

### New, and unrelated to this PR: a test that will flake in CI

`src/lib/eslintImportCoverage.spec.ts` timed out in my full local run:

```
FAIL  src/lib/eslintImportCoverage.spec.ts > reports a missing default export
Error: Test timed out in 5000ms.
```

It is not caused by this branch — the file is untouched here and it passed in my previous full run of the same branch. It is load-dependent:

- run in isolation: **1.7 s / 2.0 s / 2.7 s** across three runs, all passing
- run inside the full suite (139 files in parallel): exceeds the default 5000 ms

The test builds an ESLint instance with the type-aware config, four times, once per table case. Under parallel load the startup alone eats the budget, and there is no `testTimeout` configured in `vitest.config.ts`, so the 5 s default applies.

This machine is faster than the CI runner, so CI will hit it — intermittently, which is the worst kind. Two ways out, either is fine:

1. an explicit `{ timeout: 30_000 }` on the `it.each`, or
2. one ESLint instance created once for the whole `describe` instead of one per case, which also cuts the runtime by roughly four.

Happy to open it as its own issue if you would rather not grow this PR.

---

For completeness: the attachment upload question in the other thread is not blocking from my side.

Verified locally on Node 24.19.0 / npm 12.0.2 — `typecheck` and `lint --max-warnings=0` clean, **139 files / 854 tests** with the single flake above, PWA / Android / Electron builds all pass.

dev-adamant-im (commented):
## What should be fixed before merge

**Medium:** `getChatRooms` seeds the account-wide polling cursor from a raw room `lastTransaction.height`, including AIP-6 signal messages that are then filtered out of `messages`. `loadChats` commits that value via `setHeight`, and `getNewMessages` continues with `getChats(state.lastMessageHeight)` — so visible chat transactions below a trailing signal can be skipped until the user opens that room.

Please restore cursor semantics to **visible / successfully processed** transactions only (details on the inline comment). The `fetchedCount` / `getChats` raw-page pagination path is fine and should stay as-is.

Follow-ups already look good (`Number(type)` for signals, decode hex normalization, `fetchedCount = 0` default, signed Android checklist).

src/lib/adamant-api/index.js
```diff
@@ -770,8 +775,14 @@ export async function getChatRooms(address, params) {
     ...defaultParams,
     ...params
   })
+  const fetchedCount = chats.length
+  const lastMessageHeight = chats[0]?.lastTransaction?.height || 0
 
   const messages = chats.flatMap((chat) => {
+    if (!isChatTransactionVisible(chat.lastTransaction)) {
+      return []
```
dev-adamant-im, line 782: **Medium — fix before merge: polling cursor vs signal filter**

`lastMessageHeight` is taken from the **raw** first room:

```js
const lastMessageHeight = chats[0]?.lastTransaction?.height || 0
```

Then `lastTransaction` may be dropped by `isChatTransactionVisible` (and also by key-binding / decode failures below). Previously the cursor came from the first **visible** decoded message:

```js
const lastMessageHeight = (messages[0] && messages[0].height) || 0
```

### Why it breaks

1. `loadChats` does `commit('setHeight', lastMessageHeight)`
2. `getNewMessages` calls `getChats(state.lastMessageHeight)`
3. If the newest room activity is an AIP-6 signal at height `H`, the global cursor jumps to `H`
4. Visible chat txs below `H` (including an earlier message in that same room that the chatrooms API no longer returns as `lastTransaction`) are never pulled by account-wide polling

This can happen on a honest network whenever a signal is the latest activity — not only with a malicious node. List preview, unread counts, and early `botCommands` handling can lag until `getChatRoomMessages` runs for that contact.

### Suggested fix

Compute the cursor **after** filtering, from processed visible txs only, e.g.:

```js
const lastMessageHeight = messages.reduce(
  (max, tx) => Math.max(max, tx.height ?? 0),
  0
)
```

or take the max height among `lastTransaction` values that pass `isChatTransactionVisible` (and were actually decoded into `messages`).

Do **not** change the `getChats` / `fetchedCount` / `lastProcessedHeight` pagination for the account-wide chat stream — advancing past hidden protocol messages there is correct. The bug is specific to using a hidden chatrooms `lastTransaction` as the **sync watermark** for `getNewMessages`.

A regression test with rooms ordered `[signal@H=1000, visible@H=800]` asserting the returned `lastMessageHeight === 800` (or max visible) would lock this in.

adamantmm (approved):

src/lib/adamant-api/index.js
adamant-al: Fixed: `getChatRooms` now derives `lastMessageHeight` from the highest successfully processed visible room transaction. A hidden signal, rejected key, or decode failure can no longer advance the account-wide polling watermark. Added the requested regression with rooms ordered as `signal@1000` and `visible@800`; it returns watermark `800` while preserving raw `fetchedCount === 2`.

adamant-al:
Follow-up on the load-dependent `eslintImportCoverage` timeout reported in https://github.com/Adamant-im/adamant-im/pull/966#issuecomment-5245469398: the spec already reuses one module-level `ESLint` instance, so there was no repeated constructor to remove. I added an explicit 30-second timeout to all four type-aware lint probes instead. The complete suite now passes under parallel load: 139 files, 855 tests; lint, typecheck, production build, and `git diff --check` are also clean.

metalisk (approved):

adamant-al:
Addressed one more initial chat-list pagination edge case: `loadChats` now passes the actual `perPage` limit to `getChatRooms`, advances its offset from raw `fetchedCount`, and updates the visible-message watermark independently. A first page containing only hidden signals therefore keeps `lastMessageHeight` at `0` but advances the list offset to the raw room count instead of leaving it at `0`. Added a dedicated regression test. Full validation passes: lint, typecheck, 139 files / 856 tests, production build, and `git diff --check`.

Issue #959, opened 2026-08-08
[Task] Local verification of node-supplied data: keys, realtime events, value-bearing transactions


adamantmm:
### Summary

Define and implement what the PWA must verify locally about data received from an ADAMANT node, so that a compromised, malicious or intercepted node cannot mislead the user — without adding per-message network round trips or slowing down chat rendering.

#### Why this is separate

ADAMANT is deliberately zero-trust towards nodes: users can run their own, connect to bare IPs, to localhost, or over Tor, and payloads are encrypted client-side. That design choice is only fully realised if the client independently checks what it can. Today it largely does not: node responses are accepted as-is for public keys, for realtime events, and for value-bearing transactions.

The threat is real but currently low-probability, and the naive mitigations (polling a second node, re-fetching every transaction) would cost a request per address or per message and would degrade the chat experience. So this needs to be designed properly rather than patched in passing.

Companion work on the node side: Adamant-im/adamant#203 — an AIP defining exactly which transaction fields are covered by the signature and the transaction ID, and which are node-derived context. This issue is the client-side counterpart for the PWA. An equivalent issue should follow for the iOS client.

#### What a hostile node can and cannot do

It cannot break the cryptography: it cannot forge an Ed25519 signature, and it cannot decrypt or forge a NaCl `box` payload without the corresponding private key.

It can, unless the client checks:

1. return a public key of its own choosing for an address, so that outgoing messages are encrypted to it instead of the intended recipient
2. push realtime events for transactions that never existed
3. fabricate or alter value-bearing transactions in the fields that are not covered by the signature, or that the client never re-derives
4. omit or reorder transactions, and report false confirmation counts or heights
5. observe metadata: which addresses a client asks about, and when

#### Threat 1 — public key substitution

**This one is fully solvable offline and has already been closed on the working branch.**

An ADM address is derived from the public key: the first eight bytes of `sha256(publicKey)`, reversed, read as a decimal number, prefixed with `U`. So the address commits to the key. A node cannot supply a different key for a given address without also changing the address, and producing a second key that yields the same address is a 64-bit preimage problem.

The client therefore verifies the binding locally — one hash, no request — everywhere a key arrives: `getPublicKey`, the bulk `/api/chatrooms/*` responses that are the app's main source of keys, and the WebSocket payload. Cross-checking against a second node was considered and rejected: it costs a request per address, and a second node can be compromised too, whereas arithmetic cannot.

[...]

PR #984, opened 2026-09-13
Fix: Sync incoming ADM transfers and wallet source channels

### LINKED ISSUE #982: [Bug, UX/UI] Refresh incoming ADM transfer status in chat previews and space the status icon
## Summary

A newly received ADM transfer can remain marked with the pending hourglass in the chat list even after the transaction is registered or confirmed. The status icon also has no visual spacing from the preceding preview content.

## Steps to reproduce

1. Create or unlock a fresh ADAMANT account
2. Request free ADM from the Free Tokens service
3. Return to the chat list and wait for the incoming ADM transaction to appear
4. Wait for the transaction to become registered or confirmed
5. Optionally open the chat or transaction details and return to the chat list
6. Observe the transfer status and the spacing before its status icon in the chat preview

## Expected behavior

The chat preview should refresh the ADM transaction status automatically and stay consistent with the chat bubble, transaction list, and transaction details. The status icon should have a clear token-based gap from the preceding content.

## Actual behavior

The chat preview can keep showing the pending hourglass indefinitely. The status icon is rendered directly against the preceding content without a visible gap.

## Screenshots or videos

Not provided.

## Environment

Observed in the PWA. The exact browser and operating system were not recorded.

## Related issues

- #900 tracks the broader transaction-status synchronization work
- #947 previously addressed related transfer-status regressions and is closed

### LINKED ISSUE #983: [Task] Select adamant-wallets source branch by PWA channel
### Summary

Make `npm run wallets:data:generate` select the `adamant-wallets` source branch from the current PWA release channel: production from `master`, development from `dev`

### Details

The generator currently defaults to `adamant-wallets/dev` for every PWA branch unless a branch argument is passed manually. This can make a production update accidentally consume development wallet metadata

Use one coherent `adamant-wallets` snapshot per generator run because the generated coin data, icons, schema source, and network configs are shared artifacts in the working tree. Do not combine files from `master` and `dev` within one run

Required channel mapping:

- PWA `master` -> `adamant-wallets/master`
- PWA `dev` and topic branches -> `adamant-wallets/dev`
- An explicit `dev` or `master` argument overrides automatic selection for maintenance workflows
- Unsupported explicit source branches fail before the submodule is changed

Normal development and build commands must remain reproducible and offline with respect to `adamant-wallets`:

- `npm run dev` and `npm run build` continue to use committed generated JSON and assets without fetching or regenerating wallet metadata
- `npm run dev` continues to start the local Vite development server on `localhost:8080`; it does not trigger or deploy the remote `pwa-dev` channel

### Checklist

- [ ] Resolve the wallet source branch from the current PWA branch
- [ ] Keep explicit `dev` and `master` overrides
- [ ] Reject unsupported explicit wallet source branches before updating the submodule
- [ ] Log both the detected PWA branch and selected wallet source branch
- [ ] Document the channel mapping and the fact that regular dev/build commands use committed artifacts
- [ ] Add unit tests for production, development, topic-branch, detached-HEAD, explicit-override, and invalid-override cases

### Notes

Related issues:

- #972 tracks isolation and drift checks for generated network configurations
- #982 tracks incoming ADM transfer synchronization and chat preview status UX
- Adamant-im/adamant-wallets#143 tracks raising the minimum supported ADAMANT Node API version in canonical wallet metadata

### Verification

- [ ] Run the branch resolver unit tests
- [ ] On PWA `master`, verify that the generator selects `adamant-wallets/master`
- [ ] On PWA `dev` or a topic branch, verify that the generator selects `adamant-wallets/dev`
- [ ] Verify that `npm run wallets:data:generate -- master` and `-- dev` select the explicit source
- [ ] Verify that an unsupported explicit branch exits before the submodule is updated
- [ ] `npm run lint`
- [ ] `npm run typecheck`
- [ ] `npm run test -- --run`
- [ ] `npm run build`

al-onyxprotocol:
## Description

- Include direct ADM transfers in chat, chatroom, and WebSocket retrieval paths
- Connect the chat socket immediately with the current ADM nodes and subscribe to transaction types `0` and `8`
- Insert live and reconciled incoming transfers through an explicit unread path without double-counting duplicate transactions
- Derive direct ADM transfer confirmation status when node responses omit it
- Apply logical spacing to leading and trailing transaction status icons in chat previews
- Select `adamant-wallets/dev` for PWA development and topic branches and `adamant-wallets/master` for the PWA production branch
- Regenerate wallet data, configs, types, and the submodule reference from the current `adamant-wallets/dev` metadata

## Related issue

Closes #982

Closes #983

Related to Adamant-im/adamant-wallets#143

## External links (optional)

- [ADAMANT WebSocket subscriptions](https://docs.adamant.im/api/websocket.html)
- [ADAMANT chatrooms API](https://docs.adamant.im/api-endpoints/chatrooms.html)

Not included. Stable icon-spacing and unread behavior are covered by component, store, and socket contract tests

## Breaking changes

- ADM nodes below API version `0.10.0` remain configured but are excluded from all ADM requests,
  including balances, transfers, KVS, chats, and sockets. At the audited 2026-09-13 state, 5 of 13
  mainnet, 3 of 7 Tor, and 2 of 3 testnet endpoints satisfy this requirement
- `adamant-wallets/master` must receive the matching minimum-version metadata before the PWA
  production release; otherwise production-channel regeneration would restore the older minimum
- No backward-incompatible persisted-state changes

## How to test

1. Run `npm run dev`, sign in with a new test account, and request free ADM tokens
2. Verify that the Bounty transfer appears in the chat list without waiting for the next polling interval
3. Verify that the Bounty chat receives an unread badge and that opening the chat clears it
4. Verify that the direct transfer starts as registered and advances to confirmed
5. Verify spacing between incoming and outgoing transfer text and their status icons
6. Disable or interrupt the socket and verify that polling reconciles the transfer without duplicating it or its unread count
7. Run `npm run wallets:data:generate` on a development or topic branch and verify that it selects `adamant-wallets/dev`
8. Verify that PWA `master` selects `adamant-wallets/master`, or use `npm run wallets:data:generate -- master` for an explicit maintenance check

Validation completed:

- `npm run wallets:generate`
- `npm run lint`
- `npm run typecheck`
- `npm run test -- --run` — 141 test files and 879 tests passed
- `npm run build`
- Local Vite smoke test on `localhost:8080` with no browser console errors
- `git diff --check origin/dev...HEAD`

## Notes for reviewers

- Direct type `0` transfers bypass public-key resolution because they have no encrypted chat payload; type `8` messages retain address-to-key binding validation
- WebSocket `newTrans` payloads are provisional, so the existing REST polling path remains the durable status reconciliation mechanism
- Initial chat history keeps height-based unread restoration, while socket and post-initialization polling paths mark incoming messages explicitly
- The committed development-channel wallet snapshot uses `minVersion: 0.10.0` from `adamant-wallets/dev`; generation on PWA `master` selects the current wallet `master` metadata instead
- Regular `npm run dev` and `npm run build` commands continue to consume committed generated artifacts and do not fetch `adamant-wallets`

## Checklist

- [x] Code is formatted
- [x] Tests added/updated
- [x] Documentation updated
- [x] Changes tested in all relevant environments


src/store/plugins/socketsPlugin.js
```diff
@@ -8,6 +9,11 @@ function subscribe(store) {
   socketClient.subscribe('newMessage', (transaction) => {
     if (!isChatTransactionVisible(transaction)) return
 
+    if (transaction.type === Transactions.SEND) {
+      store.dispatch('chat/pushNewMessages', [transaction])
+      return
+    }
```
dev-adamant-im, line 55: **[High, security, related to #959]** A realtime `type 0` payload is inserted into the chat as-is and marked unread.

Dropping the key-binding step for transfers is fine by itself: a sender's public key is public, so that step never protected transfers. The problem is what happens after insertion, combined with the other changes in this PR:

- `useAdmTransactionQuery` now returns `CONFIRMED` for the on-chain record with the same `id` (see the comment there), and ADM is exempt from `getInconsistentStatus`. A realtime event that reuses the `id` of an existing transaction but carries a different `amount`, `senderId` or `message` is therefore rendered with a green check in the chat list and in the bubble. Before this PR such a socket-only transfer stayed `REGISTERED`.
- Polling reconciliation keeps the socket-supplied `amount` and `message` (see the comment on `pushMessage`).

#959 lists "fabricated realtime events do not reach the chat" as a verification goal, and its option C (re-fetch socket-delivered transfers by id) targets exactly this path.

Suggested minimum for this PR: treat the event as a trigger. Fetch the transaction by `id` via REST (`getTransaction(id)`, already used by the status query) and push the fetched record. Direct transfers are rare, so the request budget for chat messages is not affected. If that is out of scope, keep socket-only transfers visibly unverified until the REST record matches.

src/hooks/queries/transaction/useAdmTransactionQuery.ts
```diff
@@ -5,17 +5,22 @@ import { refetchIntervalFactory, refetchOnMountFn, retryDelayFactory, retryFacto
...
 
 const fetchTransaction = async (transactionId: string, currentUserAdmAddress: string) => {
   const rawTransaction = await admApi.getTransaction(transactionId, 1)
   if (!rawTransaction) throw new Error('Transaction not found')
 
   const transaction = decodeTransaction(rawTransaction, currentUserAdmAddress)
+  const status =
+    transaction.height || transaction.confirmations > 0
+      ? TransactionStatus.CONFIRMED
+      : TransactionStatus.REGISTERED
```
dev-adamant-im: **[High, security, related to #959]** The derived status is correct for the on-chain record, but it is now applied to whichever chat message carries this `id`.

`TransactionStatusProvider` queries ADM transfers by `id` only, and `useInconsistentStatusState` / `getInconsistentStatus` return `''` for ADM. The `amount`, `senderId` and `recipientId` of the chat message are therefore never compared with this record. Before this PR the query result had no `status`, so the message kept its own status, and a transfer known only from a realtime event stayed `REGISTERED`. Now it becomes `CONFIRMED`.

Please compare the fetched record with the known chat message (`knownAdmTransaction`: `senderId`, `recipientId`, `amount`) and report `INVALID` on a mismatch, instead of the unconditional `''` for ADM. This extends the existing consistency mechanism, as #959 proposes.

src/store/modules/chat/index.js
```diff
@@ -458,7 +458,7 @@ const mutations = {
    * Push an message to a specific chat by senderId.
    * @param {string} userId Your address
    */
-  pushMessage(state, { message, userId, unshift = false }) {
+  pushMessage(state, { message, userId, unshift = false, markAsUnread = false }) {
```
dev-adamant-im, line 470: **[Medium, correctness, related to #959]** Reconciliation keeps the first-seen payload.

When the same `id` is already in the chat, the dedupe branch below updates only `status` and `height`, then returns. Direct transfers now arrive via the socket first and via polling later, so the socket-supplied `amount` and `message` survive reconciliation while the status flips to `CONFIRMED`.

A local store test confirmed this. Socket payload `amount: 1e14`, then the REST record with `amount: 1e8, height: 101`: the message ends up with `amount: 1e14, status: CONFIRMED`.

Suggestion: when the incoming record comes from REST (has `height`), overwrite the value fields (`amount`, `message`) as well. `normalizeMessage` should also ignore `message` for `type 0`, because a direct transfer has no on-chain comment.

src/store/plugins/__tests__/socketsPlugin.spec.js
```diff
@@ -128,7 +163,23 @@ describe('socketsPlugin request budget', () => {
...
+      recipientPublicKey: undefined
+    })
+
+    handlers.newMessage(transfer)
+
+    expect(getPublicKey).not.toHaveBeenCalled()
+    expect(cacheVerifiedPublicKey).not.toHaveBeenCalled()
+    expect(decodeChat).not.toHaveBeenCalled()
+    expect(store.dispatch).toHaveBeenCalledWith('chat/pushNewMessages', [transfer])
   })
```
dev-adamant-im: **[Medium, tests]** This test pins the behavior described above: it asserts that the raw socket payload is dispatched unchanged.

Once the socket path fetches the record by `id`, or marks the transfer as unverified, please rewrite this test to assert that a realtime `type 0` event cannot put an unverified `amount` into the chat.

src/config/production.json
```diff
@@ -55,7 +55,7 @@
         "onScreenUpdateInterval": 10000,
         "threshold": 10
       },
-      "minVersion": "0.8.0",
+      "minVersion": "0.10.0",
```
dev-adamant-im, line 58: **[Medium, breaking change, accepted but needs to be stated]** Raising `minVersion` to `0.10.0` is an accepted trade-off (Adamant-im/adamant-wallets#143). It is still a breaking operational change, while the PR description says "No backward-incompatible client or persisted-state changes". Please list it under **Breaking changes**:

- `isActiveNode()` requires `hasMinNodeVersion()` for **all** ADM requests (balance, sending, KVS, chats), not only for the chat and socket features that need `includeDirectTransfers`.
- Per the 2026-09-13 endpoint audit in adamant-wallets#143, only 5 of 13 mainnet, 3 of 7 Tor and 2 of 3 testnet endpoints currently qualify. The rest stay listed but are unused until upgraded.
- The same applies to `development.json`, `testnet.json`, `tor.json` and `src/lib/constants/cryptos/data.json`.

Release note: `adamant-wallets/master` still has `0.8.0`, because Adamant-im/adamant-wallets#144 is only on `dev`. Promote it before the PWA release. Otherwise running `wallets:data:generate` on PWA `master` silently reverts to `0.8.0`, and nodes that ignore `includeDirectTransfers` are used again.

src/store/plugins/socketsPlugin.js
```diff
@@ -85,7 +91,7 @@ export default (store) => {
       socketClient.setNodes(store.getters['nodes/adm'])
     }
 
-    if (mutation.type === 'nodes/useFastest') {
+    if (mutation.type === 'nodes/useFastestAdmNode') {
       socketClient.setUseFastest(mutation.payload)
     }
```
dev-adamant-im, line 135: **[Low, behavior]** Now that the mutation name is fixed, `useFastestAdmNode` reaches `SocketClient`. On each 5 s revision, `reviseConnection()` in `src/lib/sockets.js` reconnects whenever the fastest node differs from the current one:

```js
(this.isOnline && this.useFastest && this.currentNode.hostname !== node.hostname)
```

The fastest-node ranking is a coarse ping estimate against `/api/node/status` (see #962), and it says nothing about a long-lived socket. Reconnecting on every ranking change only adds churn and short gaps in the subscription.

Please don't reconnect an online socket because the fastest node changed. Either drop this clause from `reviseConnection()` and keep reconnecting only when the socket is offline or its node becomes inactive, or don't wire `useFastestAdmNode` into the socket at all. The new test `tracks the ADM fastest-node setting mutation` would need to follow.

src/store/modules/chat/index.js
```diff
@@ -803,7 +824,7 @@ const actions = {
       const { messages, lastMessageHeight, nodeTimestamp } = result
       const chatsActualInterval = getters.chatsActualityTimeout
 
-      dispatch('pushMessages', messages)
+      dispatch('pushNewMessages', messages)
```
dev-adamant-im: **[Low, regression]** `pushNewMessages` bypasses the `lastMessageHeight > 0` guard on every poll.

Suppose `loadChats` finishes with `lastMessageHeight === 0` while the account already has history, for example because the first chat-room page was dropped entirely by the visibility or key-binding filters. The first poll then fetches the whole history and marks every incoming message unread. Before this PR the height guard in `pushMessage` prevented that.

A local store test with three historical incoming messages: `pushMessages` gives 0 unread, `pushNewMessages` gives 3.

Suggestion: pass `markAsUnread` only when `state.lastMessageHeight > 0`, or when the initial chat-room snapshot was genuinely empty (`fetchedCount === 0`). That way a fresh account receiving the Bounty transfer still gets the badge.

scripts/wallets/sourceBranch.mjs
```diff
@@ -0,0 +1,24 @@
...
+      throw new Error(
+        `Unsupported adamant-wallets source branch: ${requestedBranch || '(empty)'}. ` +
+          'Expected "dev" or "master".'
+      )
+    }
+
+    return requestedBranch
+  }
+
+  return pwaBranch === 'master' ? 'master' : 'dev'
```
dev-adamant-im, line 23: **[Nit, decision]** Every branch other than `master` maps to `dev`, including hotfix branches cut from `master` (for example `hotfix/*`).

That follows #983 as written, but it is exactly the case the issue wants to avoid: a production update consuming development metadata. Consider mapping `hotfix/*` to `master`, or requiring an explicit `-- master` / `-- dev` for such branches. At minimum, mention hotfix branches in the README table.

scripts/wallets.mjs
```diff
@@ -8,22 +8,33 @@ import mapValues from 'lodash-es/mapValues.js'
...
-async function run(branch = 'dev') {
+async function run(requestedBranch) {
+  const { stdout } = await $`git branch --show-current`
+  const pwaBranch = stdout.trim()
+  const branch = resolveWalletsSourceBranch({ pwaBranch, requestedBranch })
+
+  logInfo(
+    'Selecting `adamant-wallets` source:',
+    `PWA branch ${pwaBranch || '(detached HEAD)'} -> adamant-wallets/${branch}`
+  )
```
dev-adamant-im, line 44: **[Nit]** With an explicit override the log reads, for example, `PWA branch dev -> adamant-wallets/master`, which looks like a wrong mapping. Consider logging whether the source was selected automatically or by an explicit argument.

README.md
```diff
@@ -112,6 +112,31 @@ npm run dev-https
...
+| `master`           | `master`                 |
+| `dev`              | `dev`                    |
+| Any topic branch   | `dev`                    |
+| Detached HEAD      | `dev`                    |
+
+Pass `dev` or `master` after `--` to override the automatic selection for a maintenance workflow:
+
+```bash
+npm run wallets:data:generate -- master
+```
```
dev-adamant-im, line 131: **[Nit, pre-existing]** Worth noting here that the override only works with `wallets:data:generate`.

`npm run wallets:generate -- master` appends the argument to the last command in the chain. The data step runs with automatic selection, and `master` goes to `wallets:types:generate` (and ends up as an extra path for prettier) instead. Verified with a minimal npm script chain.

dev-adamant-im (commented):
## Review summary

Reviewed against #982, #983 and Adamant-im/adamant-wallets#143, including the node source for `/api/chats/get`, `/api/chatrooms` and the WebSocket subscription matching. Details are in the inline comments.

**Verdict:** not ready to merge as is. The fixes for #982 and #983 are sound, but the realtime path for direct transfers now makes unverified data look confirmed (#959), the `minVersion` bump is an undocumented breaking change, and the socket should not follow fastest-node changes.

### What's good

- #982 is fixed at the root on both layers: the ADM status query now yields a terminal status (the endless refetch is gone), and direct transfers are included in `/api/chats/get` and chat-room requests, so the store reconciles them.
- The server-side `address` + `types: [0, 8]` subscription is correct: in the node both filters must match, so the client receives only its own chat messages and transfers.
- Outgoing direct-transfer echoes no longer go through a counterparty key lookup, so they are not dropped when the recipient has no published key.
- `SocketClient.init()` clears the pending revision timer, so repeated init calls no longer stack reconnection loops.
- Explicit `markAsUnread` path with dedupe: a transfer delivered by the socket and then by polling is counted once, covered by tests.
- `isIn` -> `inId` and the removed trailing slash are safe: the node handles `isIn` and `inId` identically, and `fromHeight` is strict (`>`), so polling does not refetch the last block.
- Status icon spacing uses logical properties and design tokens; the stale `nodes/useFastest` mutation name is fixed.
- #983: one coherent snapshot per run, strict `dev`/`master` whitelist validated before the submodule is touched, a README section, and unit tests for every case listed in the issue, including detached HEAD.
- No new dependencies. The submodule bump only changes `minVersion`, CHANGELOG and OpenAPI examples. Good unit coverage for the new paths.

### What to pay attention to

- **Security (#959).** The PR does not create this class of problem, but it widens it: the realtime `type 0` payload is inserted as-is, the status lookup by `id` now turns it `CONFIRMED`, and ADM is still exempt from the consistency check. #959 sets "fabricated realtime events do not reach the chat" as a goal, and a new test pins the opposite.
- **Breaking change.** `minVersion 0.10.0` is accepted, but it has to be stated in the PR description, and `adamant-wallets/master` has to be promoted before the release.
- **Socket reconnects** on fastest-node changes are newly activated by the mutation-name fix and should not happen.
- **Scope.** One commit bundles #982, #983 and the node minimum-version bump, while #982 does not mention the unread and instant-appearance changes. Consider updating the issue or splitting out the version bump so each part can be reverted independently. Adamant-im/adamant-wallets#143 is still titled "v0.9.0" although it was closed with `0.10.0`.

### Findings

| # | Severity | Area | Finding |
|---|---|---|---|
| 1 | High | Security, #959 | Unverified realtime direct transfers can be shown as confirmed: the payload is inserted as-is (`socketsPlugin.js`) and the by-`id` `CONFIRMED` status is applied without comparing `amount` / `senderId` / `recipientId` (`useAdmTransactionQuery.ts`, ADM exempt in `getInconsistentStatus`) |
| 2 | Medium | Correctness, #959 | Reconciliation updates only `status` and `height`, so the socket-supplied `amount` and `message` survive (`pushMessage`) |
| 3 | Medium | Tests | `pushes a direct ADM transfer immediately...` locks in the raw insertion |
| 4 | Medium | Breaking change, accepted | `minVersion 0.10.0` excludes most endpoints from all ADM requests; not stated in the description; `adamant-wallets/master` still has `0.8.0` |
| 5 | Low | Behavior | The socket reconnects whenever the fastest node changes, based on the ping ranking (#962) |
| 6 | Low | Regression | The whole history is marked unread when polling starts with `lastMessageHeight === 0` |
| 7 | Nit | Tooling | Hotfix branches cut from `master` select `adamant-wallets/dev` |
| 8 | Nit | Tooling | The log does not show that an explicit override was used |
| 9 | Nit | Docs | `npm run wallets:generate -- master` does not forward the override to the data step (pre-existing) |

### CI

- Lint, typecheck, unit tests, build, CodeQL and deployments pass.
- `Verify dependency integrity and advisories` fails, but not because of this PR: the same step fails on `dev` (b378c2f8), and no manifest or lockfile is changed here.
- `Run Playwright smoke checks`: 6 failed, 103 passed. The failures look environmental. All five `send-funds-live-account` failures stop while waiting for a positive balance or transferable amount, and the live test account `U3716604363012166999` currently holds 0 ADM (checked on two nodes). `chat-composer-regression.spec.ts:153` also failed on the last `dev` run (787d3673). Worth re-running once the test account is funded.

src/store/plugins/socketsPlugin.js
al-onyxprotocol: After the follow-up UX decision, incoming value-bearing socket transactions are shown immediately, but socket-supplied confirmation metadata is ignored. Both type `0` and type `8` with `amount > 0` are forced to `REGISTERED`; REST reconciliation and the by-ID status query decide whether they become `CONFIRMED` or `INVALID`. This is an interim boundary until #959 adds local signature verification.

src/hooks/queries/transaction/useAdmTransactionQuery.ts
al-onyxprotocol: I extended the existing consistency mechanism to ADM. The fetched record is now compared with `knownAdmTransaction` by transaction ID, amount (with the raw ADM unit converted), `senderId`, and `recipientId`; a mismatch produces the existing inconsistency reason and therefore `INVALID`. ADM consistency checks also avoid unrelated KVS address lookups.

src/store/modules/chat/index.js
al-onyxprotocol: After the follow-up UX decision, duplicate reconciliation compares the first-seen ADM amount and participants with the REST record before changing status. Matching records update status, height, and confirmations; a mismatch becomes `INVALID` and is never silently upgraded to confirmed. The first-seen fields remain available to the status provider for the same REST/socket comparison, and direct type `0` normalization ignores any supplied `message` field.

src/store/plugins/__tests__/socketsPlugin.spec.js
al-onyxprotocol: After the follow-up UX decision, the socket tests now require incoming type `0` and value-bearing type `8` transactions to be dispatched immediately as `REGISTERED`, even if the event claims confirmations or a confirmed status. Store and query tests cover matching REST reconciliation, mismatched fields becoming `INVALID`, and `CONFIRMED` being allowed only at `confirmations >= 1`.

src/config/production.json
al-onyxprotocol: I updated the PR's **Breaking changes** section with the `0.10.0` requirement for all ADM requests, the audited mainnet/Tor/testnet endpoint counts, and the requirement to promote the matching metadata to `adamant-wallets/master` before the production release. It also states separately that persisted state remains compatible.

src/store/plugins/socketsPlugin.js
al-onyxprotocol: I removed fastest-ranking changes from the reconnect condition. `useFastest` still chooses a node for the initial connection and recovery, but a healthy active socket now stays connected when ping rankings change. I added a regression test with a different newly fastest node.

src/store/modules/chat/index.js
al-onyxprotocol: I now mark polled messages unread only when there is a positive height baseline or the initial snapshot was genuinely empty (`offset === 0`). If the initial page contained records that were filtered out, polling uses the history path and does not mark the recovered history unread. Both cases are covered by store tests.

scripts/wallets/sourceBranch.mjs
al-onyxprotocol: I kept the mapping required by #983: only PWA `master` selects wallet `master`, while all topic branches select `dev`. The README now calls out `hotfix/*` explicitly and instructs production hotfix maintenance to use the explicit `master` override, avoiding an implicit channel decision based only on a branch name.

scripts/wallets.mjs
al-onyxprotocol: I changed the log to label the selection as either `automatic mapping` or `explicit override`. Both forms still show the detected PWA branch and the selected `adamant-wallets` branch.

README.md
al-onyxprotocol: I documented that the override applies only to `wallets:data:generate` and that the chained `wallets:generate` command does not forward trailing arguments to its data step.

src/store/modules/chat/index.js
```diff
@@ -458,25 +459,48 @@ const mutations = {
...
+      return !!existingAdmTransfer
+    })
+    if (existingAdmTransfer) {
+      const hasMismatch =
+        message.type !== Cryptos.ADM ||
+        !isNumeric(existingAdmTransfer.amount) ||
+        !isNumeric(message.amount) ||
+        !new BigNumber(existingAdmTransfer.amount).isEqualTo(message.amount) ||
+        !isStringEqualCI(existingAdmTransfer.senderId, message.senderId) ||
+        !isStringEqualCI(existingAdmTransfer.recipientId, message.recipientId)
```
dev-adamant-im, line 491: **[High, regression]** The user's own in-chat ADM transfers become `INVALID` after polling.

`SendFundsForm.pushTransactionToChat()` stores the optimistic chat record as `amount * 1e8`. For many amounts that is not an integer: `0.29 * 1e8 = 28999999.999999996`, and the same happens for `0.57`, `1.1` and `2.675`. The transaction that is actually broadcast goes through `adamant.prepareAmount()` (`Math.round`) in `sendMessage()` / `sendTokens()`, so REST returns `29000000`.

This exact comparison then fails. The record turns `INVALID` with "Amount mismatch — possible fraud" and never recovers, because the first-seen fields are kept. Through `getInconsistentStatus()` the same record also marks the transaction details and the ADM wallet row as invalid.

A local test against this mutation and `getInconsistentStatus()`: 0.29, 0.57 and 1.1 ADM end up `INVALID` / `wrong_amount`; 5 ADM ends up `CONFIRMED`.

The exact comparison itself is right. On-chain ADM amounts are always integers (the node rejects non-integer amounts with "Invalid transaction amount"), so incoming records from the socket and from REST carry the same integer. Only the optimistic local record is off. Please store `adamant.prepareAmount(this.amount)` in `pushTransactionToChat()`, which is exactly the value that is broadcast, and add a test that reconciles a locally pushed 0.29 ADM transfer with its REST record.

src/components/transactions/utils/getInconsistentStatus.ts
```diff
@@ -28,10 +31,28 @@ export function getInconsistentStatus(
...
+      return TransactionInconsistentReason.WRONG_TX_HASH
+    }
+
+    if (
+      !isNumeric(transaction.amount) ||
+      !isNumeric(admTransaction.amount) ||
+      !new BigNumber(transaction.amount).times(AdmAmountMultiplier).isEqualTo(admTransaction.amount)
+    ) {
+      return TransactionInconsistentReason.WRONG_AMOUNT
+    }
```
dev-adamant-im, line 45: **[High, same root cause as the comment on `pushMessage`]** This is where the non-integer optimistic record shows up as `wrong_amount` in the transaction details and in the ADM wallet row.

Please keep this comparison exact rather than moving ADM to the 0.3% tolerance used for coins. For coins, the message amount is a decimal string written by the sender's client and compared with another chain's precision. For native ADM, both sides describe the same on-chain integer, so any difference is a different claim. A 0.3% tolerance would let a provisional record be inflated by 3,000 ADM per 1,000,000 ADM and still turn green.

src/store/plugins/socketsPlugin.js
```diff
@@ -1,14 +1,59 @@
...
+      ![Transactions.SEND, Transactions.CHAT_MESSAGE].includes(transaction.type)
+    )
+      return
     if (!isChatTransactionVisible(transaction)) return
+    if (transaction.type === Transactions.SEND && !(Number(transaction.amount) > 0)) return
+
+    const accountAddress = store.state.address
+    const isIncoming = isStringEqualCI(transaction.recipientId, accountAddress)
+    if (!isIncoming && !isStringEqualCI(transaction.senderId, accountAddress)) return
+    const isIncomingAdmTransfer = isIncoming && Number(transaction.amount) > 0
```
dev-adamant-im: **[Low]** Only incoming value-bearing events are made provisional.

An outgoing echo, for example a transfer sent from another device, keeps the socket-supplied `confirmations`, `height` and `status`. `normalizeMessage()` then stores it as `CONFIRMED` whenever the payload claims `confirmations >= 1`, and it stays that way until the REST status check replaces it.

The rule "socket payloads never supply confirmation data" is simpler and safer if it applies in both directions: drop the `isIncoming` condition for value-bearing transactions.

src/store/modules/chat/index.js
```diff
@@ -458,25 +459,48 @@ const mutations = {
...
 
+    // Match native ADM transfers by ID across chats: a changed sender/recipient must be reported
+    // against the original record, not inserted into a second chat as an unrelated transfer.
+    let existingAdmTransfer
+    Object.values(state.chats).some((chat) => {
+      existingAdmTransfer = chat.messages.find(
+        (localMessage) => localMessage.type === Cryptos.ADM && localMessage.id === message.id
+      )
+      return !!existingAdmTransfer
+    })
```
dev-adamant-im, line 483: **[Low, performance]** This walks every message in every chat for each pushed message: plain text messages, history pages and every polling batch included.

With long loaded histories that is O(pushed messages × loaded messages) on the path that #959 asks to keep free of rendering and scrolling regressions. Consider an id index for ADM transfers, for example a `Map` from transaction id to message that `pushMessage` and `reset` maintain, so this lookup stays O(1).

src/components/TransactionListItem.vue
```diff
@@ -136,6 +147,25 @@ export default {
     } = useTransactionQuery(toRef(props, 'id'), props.crypto, {
       enabled: hasLiveStatusTracking
     })
+    const knownAdmTransaction =
+      props.crypto === Cryptos.ADM ? useFindAdmTransaction(toRef(props, 'id')) : ref(undefined)
```
dev-adamant-im: **[Low, performance]** Every ADM row now calls `useFindAdmTransaction()`, which walks the messages of all chats.

For rows without a chat record the computed scans everything, and it re-runs whenever a chat message array changes, such as on each poll or socket message. A long wallet list therefore costs rows × loaded messages per chat update. The same id index suggested for `pushMessage` would turn this into a lookup.

src/store/modules/chat/index.js
```diff
@@ -458,25 +459,48 @@ const mutations = {
...
+    if (existingAdmTransfer) {
+      const hasMismatch =
+        message.type !== Cryptos.ADM ||
+        !isNumeric(existingAdmTransfer.amount) ||
+        !isNumeric(message.amount) ||
+        !new BigNumber(existingAdmTransfer.amount).isEqualTo(message.amount) ||
+        !isStringEqualCI(existingAdmTransfer.senderId, message.senderId) ||
+        !isStringEqualCI(existingAdmTransfer.recipientId, message.recipientId)
+
+      existingAdmTransfer.status = hasMismatch ? TS.INVALID : message.status
```
dev-adamant-im, line 493: **[Low, related to #959]** Because the first-seen record is kept, whoever delivers a transfer id first decides what REST is compared against.

A realtime event that carries the id of a real incoming payment with a different amount or sender makes that real payment `INVALID` permanently: every later REST record is compared with the same socket-supplied fields. Combined with the non-integer amount issue above, a false `INVALID` on the user's own transfer is permanent too.

Consider marking socket-inserted transfers as provisional. When REST disagrees with a provisional record, replace it with the REST record, inserted through the normal path and into the correct chat, instead of pinning the chat to the socket's version.

src/store/plugins/socketsPlugin.js
```diff
@@ -1,14 +1,59 @@
...
+      if (
+        Number(transaction.amount) > 0 &&
+        Object.values(store.state.chat.chats || {}).some((chat) =>
+          chat.messages?.some((message) => message.id === transaction.id)
+        )
+      ) {
+        return
+      }
+
+      store.dispatch('chat/pushNewMessages', [asProvisionalTransaction(decodedTransaction)])
```
dev-adamant-im, line 46: **[Info, pre-existing, #959]** Not introduced here, but relevant to this interim boundary. A provisional incoming transfer still becomes `chat/lastUnreadMessage`, so while the tab is in the background the push notification built in `src/lib/notifications.js` shows "Received X ADM" with the sender's display name, using the unverified socket amount. The notification stays in the OS notification center even if the transfer later becomes `INVALID` or `REJECTED`.

One option within this PR's model: delay notifications for value-bearing transfers until their status leaves `REGISTERED`, or leave the amount out of the notification text.

src/store/modules/chat/index.js
al-onyxprotocol: Fixed. `SendFundsForm.pushTransactionToChat()` now stores the optimistic ADM record as `adamant.prepareAmount(this.amount)`, the same integer that `sendMessage()` and `sendTokens()` broadcast, so this exact comparison matches for amounts such as 0.29, 0.57 and 1.1 ADM. Added `SendFundsForm` tests for these amounts that expect integer units (`29000000`, `57000000`, `110000000`).

src/components/transactions/utils/getInconsistentStatus.ts
al-onyxprotocol: Kept this comparison exact. The non-integer value came from the optimistic chat record, which now stores integer units (see the thread on `pushMessage`), so the transaction details and the ADM wallet row compare equal values.

src/store/plugins/socketsPlugin.js
al-onyxprotocol: Fixed. Every value-bearing socket transaction is now provisional, incoming or outgoing: the socket-supplied `confirmations`, `height` and `status` are replaced with `0`, `0` and `REGISTERED` before the record reaches the store. Added tests for outgoing type `0` and type `8` echoes that claim confirmations.

src/store/modules/chat/index.js
al-onyxprotocol: Leaving this scan as is. Measured with the real mutation on reactive state: 25 pushed messages take about 7 ms with 2,000 loaded messages and about 70 ms with 30,000. The scan runs inside the mutation without dependency tracking, and it is what lets reconciliation find a transfer that was first seen in another chat. An id index would have to stay consistent with state restore and reset, which is more risk than this cost justifies. The per-row lookup in `TransactionListItem` was the actual hotspot and is fixed in that thread.

src/components/TransactionListItem.vue
al-onyxprotocol: Fixed. ADM wallet rows now look for the chat record only in the chats of the transaction's sender and recipient: `useFindAdmTransaction()` accepts several preferred chats and a `searchAllChats: false` option. `useInconsistentStatusState()` no longer falls back to scanning every chat when the caller passes its own record lookup, which would otherwise undo the scoped search. Measured with the real hook: 50 rows with 2,000 loaded messages went from about 67 ms to 0.07 ms per new chat message, and 100 rows with 30,000 loaded messages take 0.14 ms. Added tests for the scoped lookup, for the skipped fallback and for the lookup arguments of the wallet row.

src/store/modules/chat/index.js
al-onyxprotocol: Keeping the first-seen comparison in this PR. Now that optimistic records use integer amounts, a permanent `INVALID` on a legitimate transfer requires a socket node that delivers the id of a real incoming payment with different fields before the honest record arrives. The result is a visible inconsistency warning, not a false confirmation. Replacing provisional records with REST records would mean moving records between chats and reversing the UX decision made in this PR. The clean fix is the local verification planned in #959, which can verify or discard a provisional record before it is shown.

src/store/plugins/socketsPlugin.js
al-onyxprotocol: Not changing notifications in this PR. The behavior predates it, and #959 already includes "Define the UI treatment for unverified and for failed-verification transactions", which covers what a notification may show for a provisional transfer.

adamantmm (approved):

PR #986, opened 2026-09-15
Fix: Keep generated network configs isolated across build modes

### LINKED ISSUE #972: [Bug] Keep generated network configs isolated across build modes
### Summary

Generated network configuration does not consistently apply environment overrides from the canonical `adamant-wallets` metadata. The current testnet build uses mainnet IPFS nodes, mainnet ADM explorer URLs, and the mainnet ADM minimum node version

The build-target audit also confirmed the intended Tor contracts: Tor nodes and services use onion endpoints, third-party coin explorers remain clearnet, `tor-testnet` is intentionally unsupported, and the supported `tor-dev` deployment builds the mainnet Tor configuration from the `dev` branch on an external server rather than in GitHub Actions

During implementation it also turned out that every build bundled all generated network configuration files and selected one at runtime, and that `development.json` and `production.json` were always generated from the same base metadata

### Steps to reproduce

1. Compare `adamant-wallets/assets/general/adamant/info.json` with `src/config/testnet.json`
2. Inspect `updateTestnetConfig()` and `updateConfig()` in `scripts/wallets.mjs`
3. Run `npm run build:testnet`
4. Open `Options -> Nodes` in the resulting build or inspect the compiled runtime configuration

The canonical testnet metadata contains:

```json
{
  "services": {
    "ipfsNode": {
      "list": [
        { "url": "https://ipfs1test.adamant.im" },
        { "url": "https://ipfs2test.adamant.im" }
      ]
    }
  }
}
```

The committed `src/config/testnet.json` and the deployed master artifact instead contain `ipfs4.adm.im`, `ipfs6.adamant.business`, and `ipfs4.bbry.org`

### Expected behavior

Every supported build target must bundle exactly one network configuration, resolve it deterministically from the pinned `adamant-wallets` metadata, apply the complete environment override without mutating the base configuration, and fail at build time for unsupported modes

The `development` and `production` Vite modes and unit tests must use one mainnet configuration. Testnet must use all fields from the canonical `testnet` override. Tor and `tor-dev` must use the mainnet Tor configuration with onion node and service endpoints; clearnet third-party explorer links are intentional. `tor-testnet` must remain explicitly unsupported

### Actual behavior

Audit snapshot against the current `adamant-wallets` submodule:

| Target | Effective config | Result |
| --- | --- | --- |
| Development PWA | `development.json` | Matches base metadata |
| Production PWA | `production.json` | Matches base metadata |
| Testnet PWA | `testnet.json` | Five mismatches |
| Tor PWA | `tor.json` | Matches mainnet Tor metadata |
| Tor dev deployment | `tor.json` from `dev` | Supported server-side auto-build outside GitHub Actions |
| Tor testnet | None | Intentionally unsupported |
| Electron | Production | No separate testnet or Tor target |
| Capacitor Android | Production | No separate testnet or Tor target |

The five testnet mismatches are:

- `adm.explorer` uses `https://explorer.adamant.im` instead of `https://testnet.adamant.im`
- `adm.explorerTx` uses the mainnet explorer
- `adm.explorerAddress` uses the mainnet explorer
- `adm.nodes.minVersion` is `0.8.0` instead of the testnet value `0.9.0`. At the pinned revision `d2e1b88`, both values are `0.10.0`, so this field no longer differs, but the testnet value must still come from the `testnet` override.
- `adm.services.ipfsNode.list` uses mainnet IPFS nodes instead of `ipfs1test.adamant.im` and `ipfs2test.adamant.im`

Every build bundled `development.json`, `production.json`, `testnet.json`, and `tor.json`, so testnet and Tor bundles also carried the mainnet runtime configuration. `development.json` and `production.json` had identical content

Tor node and service endpoints are onion addresses. The 12 clearnet explorer URL fields inherited for BTC, DASH, DOGE, and ETH are intentional user-navigation targets and are not network configuration defects

`tor-testnet` must not be produced by merging the existing `testnet` and `tor` objects because the Tor override describes mainnet coin nodes. Unsupported mode handling should reject this combination clearly rather than silently composing incompatible networks

The supported `tor-dev` deployment is available at http://il2fcw65jrrv4au7mtuvodzfwtc6hje6r4ooz7yoqu2jl422b44iofyd.onion and is automatically built from `dev` by external server automation, not by a GitHub Actions workflow

The external Tor builds for `dev` and `master` did not use `--mode tor`. Their cron scripts copied `tor.json` over `production.json` and ran `npm run build`, so any change that stops reading `production.json` would silently ship the mainnet clearnet configuration to the onion sites

### Root cause

`updateTestnetConfig()` only replaces `config.nodes.list`:

```js
if (config.testnet) config.nodes.list = config.testnet.nodes.list
```

It ignores testnet explorer fields, the rest of the ADM node object, and testnet services. `updateConfig()` then writes `config.explorer*`, `config.nodes`, and `config.services` from the partially mutated base object

The generation functions also mutate shared config objects in sequence, which makes future variants order-dependent. The testnet deployment workflow runs `schema:generate` and `build:testnet` but does not regenerate or check wallet-derived configuration, so committed drift is bundled and deployed unchanged

`src/config/index.js` imports every generated configuration and selects one by `import.meta.env.MODE` at runtime, so no bundle excludes the configurations of other networks. It also returns `undefined` for unknown modes instead of rejecting them during build configuration

### Required changes

- [x] Replace mode-specific mutation with one pure variant resolver that starts from a cloned base object and applies the complete selected override
- [x] Replace arrays instead of index-merging them, preserving the current endpoint-list semantics
- [x] Replace the identical `development.json` and `production.json` with one `mainnet.json` shared by the `development` and `production` modes and unit tests
- [x] Regenerate `mainnet.json`, `testnet.json`, and `tor.json` from the pinned `adamant-wallets` revision
- [x] Correct the testnet mismatches listed above
- [x] Bundle exactly one network configuration for each build target and mode
- [x] Reject JSON files in `src/config` that the generator does not create, so a deployment script cannot replace a network configuration
- [x] Document the supported build-mode matrix for PWA, Electron, and Capacitor
- [x] Document `tor-dev` as the external server auto-build of the mainnet Tor configuration from `dev`
- [x] Switch the external Tor build scripts for `dev` and `master` from copying `tor.json` over `production.json` to `npm run build:tor`, and copy the build output only after a successful build
- [x] Keep `tor-testnet` explicitly unsupported and fail with a clear build-time error if that or another unknown mode is requested
- [x] Preserve the Tor contract: node and service endpoints remain onion, while BTC, DASH, DOGE, and ETH explorer links may remain clearnet
- [x] Add a non-mutating `wallets:data:check` command that fails when committed generated files differ from the pinned metadata
- [x] Run the drift check in the Quality workflow and in the GitHub Actions workflows that deploy or package the app. Deployments outside GitHub Actions, including Vercel and `tor-dev`, do not repeat it because drift depends only on the commit, while mode validation and the bundle gate run inside every Vite build.
- [x] Add tests covering complete overrides, array replacement, source immutability, supported modes, and unsupported-mode rejection
- [x] Add bundle-level assertions that forbid mainnet ADM/IPFS endpoints in testnet runtime configuration and forbid clearnet node/service endpoints in Tor runtime configuration without rejecting intentional clearnet explorer links
- [x] Document the relationship between `src/lib/constants/cryptos/data.json`, `src/config/*.json`, the pinned submodule revision, branches, deployment paths, and network modes
- [x] Move developer documentation from `README.md` to `CONTRIBUTING.md` and keep only user-facing information in `README.md`

### Verification

- [x] `npm run wallets:data:check`
- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm run test -- --run`
- [x] `npm run build`
- [x] `npm run build:testnet`
- [x] `npm run build:tor`
- [x] `npm run electron:build:prepare`
- [x] `npm run android:prebuild`
- [x] Verify the external Tor build automation for `dev` and `master` and switch it to `npm run build:tor` with the mainnet Tor configuration
- [ ] Confirm the first external `tor-dev` build with `npm run build:tor` after this change reaches `dev`
- [x] Inspect generated bundles for target-specific endpoint allowlists and forbidden cross-network endpoints

### Environment

- Repository: `Adamant-im/adamant-im`
- Audited branches: `dev` and `master`
- Master commit deployed by the latest successful testnet workflow: `dfe094d34f28a0be8e563c87271a849758ca0545`
- `adamant-wallets` revision at the audit: `9d5c4c599ea89327010930dee8cb082c2f3a9f08`
- Pinned `adamant-wallets` revision used for regeneration: `d2e1b88919f059b0278cda1fb717664602852df2`
- Affected testnet deployment: http://msg-testnet.adamant.im
- Supported external `tor-dev` deployment: http://il2fcw65jrrv4au7mtuvodzfwtc6hje6r4ooz7yoqu2jl422b44iofyd.onion
- Successful workflow containing the incorrect testnet runtime configuration: https://github.com/Adamant-im/adamant-im/actions/runs/31443555649

Related historical issues #875 and #911 updated other parts of `scripts/wallets.mjs`, but neither covered complete environment overrides, mode isolation, or generated-config drift checks

### Screenshots or videos

Not applicable; the mismatch is reproducible from committed configuration and the archived build artifact

adamant-al:
## Description

- Resolve every network configuration variant with one pure resolver that clones the base `adamant-wallets` metadata, applies the complete `testnet` or `tor` override, and replaces endpoint arrays instead of merging them by index
- Replace the identical `development.json` and `production.json` with one generated `mainnet.json` shared by the `development` and `production` modes and unit tests
- Regenerate `mainnet.json`, `testnet.json`, and `tor.json` from the pinned `adamant-wallets` revision `d2e1b88`. Testnet now uses the testnet ADM explorer and the `ipfs1test` and `ipfs2test` IPFS nodes.
- Bundle exactly one network configuration per build through `virtual:adamant-network-config` instead of bundling every generated file and selecting one at runtime
- Reject unsupported Vite modes, including `tor-testnet`, while the build configuration is resolved. Electron and Capacitor Android accept only `development` and `production`.
- Fail every Vite build before bundling when `src/config` contains a JSON file that the generator does not create, so a deployment script cannot replace a network configuration
- Fail `vite build` when a bundle includes another network configuration, a Tor node or service endpoint is not an onion address, a clearnet build contains onion endpoints, or mainnet and testnet share ADM node, IPFS node, or ADM explorer origins
- Add the non-mutating `npm run wallets:data:check` drift check and `npm run wallets:data:generate -- --pinned` for regeneration without syncing a branch
- Check out the pinned submodule and run the drift check in the Quality workflow and in the GitHub Actions workflows that deploy or package the app
- Move developer documentation from `README.md` to `CONTRIBUTING.md` and document the build mode matrix, deployment paths, and the relationship between wallet metadata, generated files, branches, and network modes
- Enable `.ts` import extensions in `jsconfig.json` so editors no longer report TS5097 in Vite config files

## Related issue

Closes #972

Not applicable; the change affects build configuration and the generated runtime network configuration

## Breaking changes

- `src/config/development.json` and `src/config/production.json` are removed. Runtime code keeps importing the network configuration through `@/config`, and the `Config` type now comes from `src/config/mainnet.json`.
- `vite build` and `vite dev` fail for unsupported modes. Electron and Capacitor Android builds no longer accept `testnet` or `tor`.
- Builds fail when `src/config` contains JSON files other than `mainnet.json`, `testnet.json`, and `tor.json`. Deployment scripts that copy `tor.json` over `production.json` must build with `npm run build:tor` instead.
- GitHub Actions workflows that deploy or package the app check out `adamant-wallets` and stop when generated files drift from the pinned revision
- No persisted user data changes

## How to test

1. Run `git submodule update --init adamant-wallets` and `npm run wallets:data:check`; the check passes
2. Change a value in `src/config/testnet.json` and run `npm run wallets:data:check` again. The check fails and reports the file; restore it with `npm run wallets:data:generate -- --pinned`.
3. Run `npm run build:testnet` and `npm run serve:testnet`, then open `Options -> Nodes`; ADM nodes are testnet nodes and IPFS nodes are `ipfs1test.adamant.im` and `ipfs2test.adamant.im`
4. Run `npm run build:tor`; the build passes the network configuration gate
5. Copy `src/config/tor.json` to `src/config/production.json` and run `npm run build:tor` again. The build fails before bundling and names `production.json`; delete the copy afterwards.
6. Run `npx vite build --config vite-pwa.config.ts --mode tor-testnet`; the build fails before bundling with an unsupported mode error
7. Run `npm run dev` and verify that `Options -> Nodes` lists mainnet nodes

Validation completed:

- `npm run wallets:data:check`
- `npm run lint`
- `npm run typecheck`
- `npm run test -- --run` — 150 test files and 1007 tests passed
- `npm run build`, `npm run build:testnet`, and `npm run build:tor`
- `npm run electron:build:prepare`
- `npm run android:prebuild`
- Bundle inspection: each bundle contains one runtime network configuration that equals its committed file. The Tor bundle has 25 of 25 onion node and service endpoints, while the mainnet and testnet bundles have none.
- The legacy server flow that copies `tor.json` over `production.json` fails before bundling
- `vite build --mode tor-testnet` fails before bundling
- `npm run dev` resolves `@/config` to `src/config/mainnet.json`
- `git diff --check`

## Notes for reviewers

- `src/lib/constants/cryptos/data.json` intentionally keeps the complete metadata snapshot for coin properties and the developer wallet metadata screen. A plain text search of any bundle therefore still finds endpoints of other networks, so the build gate checks the bundled runtime network configuration instead.
- At the pinned revision, mainnet and testnet `minVersion` are both `0.10.0`, so the regenerated testnet configuration changes only the ADM explorer fields and IPFS nodes
- Drift depends only on the commit, so Vercel and server-side builds do not run the drift check. Mode validation and the bundle gate still run inside their Vite builds.
- The external server-side Tor builds for `dev` and `master` copied `tor.json` over `production.json` and ran `npm run build`. Their cron scripts now run `npm run build:tor` and copy the output only after a successful build, which works both before and after this change.
- The server-side PWA builds for `master` and `dev` run `npm run build` without replacing configuration files and need no changes
- The generator removes JSON configurations that it no longer generates, the drift check reports them, and the build rejects them
- Electron and Android were verified with `npm run electron:build:prepare` and `npm run android:prebuild`; packaged apps were not launched
- The `Verify dependency integrity and advisories` check also fails on `dev` because of existing npm advisories in the toolchain and is unrelated to this change
- Risk areas: Tor endpoint isolation, node configuration, and CI deployment workflows

## Checklist

- [x] Code is formatted
- [x] Tests added/updated (if needed)
- [x] Documentation updated (if needed)
- [ ] Changes tested in all relevant environments

🤖 Generated with [Claude Code](https://claude.com/claude-code)


vite-config/plugins/networkConfigPlugin.ts
```diff
@@ -0,0 +1,304 @@
...
+  config: NetworkConfig,
+  references: NetworkConfigReferences
+): string[] {
+  const violations: string[] = []
+
+  for (const [coin, coinConfig] of Object.entries(config)) {
+    for (const field of EXPLORER_FIELDS) {
+      const value = coinConfig[field]
+
+      if (value !== undefined && !parseHttpUrl(value)) {
```
dev-adamant-im, line 216: **[P2] Keep the ADM explorer inside Tor**

The clearnet-explorer exception is intentionally limited to BTC, DASH, DOGE, and ETH, but this loop only validates the scheme and therefore also accepts a clearnet adm.explorer, explorerTx, or explorerAddress. I reproduced this by replacing all three ADM Tor explorer URLs with the mainnet values; findNetworkConfigViolations('tor', ...) returned no violation. A future metadata regression would make Tor users leave the onion explorer boundary when opening ADM transactions or addresses. Please require an onion host for the ADM explorer fields while preserving the explicit exemption for the four third-party coin explorers.

vite-config/plugins/networkConfigPlugin.ts
```diff
@@ -0,0 +1,304 @@
...
+
+    generateBundle(_options, bundle) {
+      if (!isBuild) return
+
+      const moduleIds = Object.values(bundle).flatMap((output) =>
+        output.type === 'chunk' ? output.moduleIds : []
+      )
+      const violations = [
+        ...findBundledNetworkConfigViolations(variant, moduleIds, configsDir),
+        ...findNetworkConfigViolations(variant, readNetworkConfig(configsDir, variant), {
```
dev-adamant-im: **[P2] Validate the emitted configuration, not the source file**

At this point the bundle check uses moduleIds, then validates a fresh read of src/config/<variant>.json; the emitted chunk code is never inspected. I reproduced a Tor chunk with the correct tor.json module ID but a clearnet ADM endpoint in output.code, and the gate accepted it. A transform or a change between load and generateBundle can therefore alter the runtime endpoints while this security gate still passes. Please bind the assertion to the serialized configuration actually emitted, for example with a verified digest or marker, or narrow the documented guarantee from emitted-bundle validation to source and module-graph validation.

vite-config/plugins/networkConfigPlugin.ts
```diff
@@ -0,0 +1,304 @@
...
+  return normalizePath(path.join(configsDir, `${variant}.json`))
+}
+
+function readNetworkConfig(configsDir: string, variant: NetworkConfigVariant): NetworkConfig {
+  // eslint-disable-next-line security/detect-non-literal-fs-filename -- variant names come from a fixed allowlist
+  return JSON.parse(readFileSync(getNetworkConfigPath(configsDir, variant), 'utf8'))
+}
+
+function collectListEndpoints(listPath: string, endpointList: EndpointList | undefined) {
+  return (endpointList?.list ?? []).flatMap((entry, index) => [
```
dev-adamant-im, line 549: **[P2] Reject missing and malformed endpoint lists**

Treating an absent list as an empty array makes the gate fail open. findNetworkConfigViolations('tor', {}), a config with missing endpoint lists, and even string-valued nodes or services all return no violations. Such a build can pass while the app has no usable nodes or services; external Tor and server builds intentionally do not run the drift check. Please validate the runtime schema and require the expected non-empty node and service lists before applying the URL and onion checks.

dev-adamant-im (changes requested):
## What looks good

- The resolver is now pure, applies the complete override, replaces endpoint arrays rather than index-merging them, and includes a prototype-pollution regression test
- mainnet.json is semantically identical to the previous development and production configs, tor.json is semantically unchanged, and testnet contains the intended ADM explorer and IPFS corrections
- The virtual module selects one runtime configuration per mode, unsupported PWA, Electron, and Android mode combinations fail before bundling, and every deploy or package workflow checks the pinned generated artifacts
- The current built outputs are isolated as intended: mainnet and testnet contain no onion endpoints, all 25 Tor node and service endpoints are v3 onion addresses, and mainnet and testnet share no ADM node, IPFS, or explorer origins

## What to watch

The current generated files and normal Vite outputs are correct. The remaining issues are fail-open gaps in the new security and reliability gate: it permits a clearnet ADM explorer in Tor, validates the source JSON rather than emitted endpoint content, and accepts empty or structurally invalid configs.

The current CI also has two red checks. The Playwright run ends with a failure in an existing transaction-scroll scenario, after an initial chat-composer failure passed on retry; I found no changed code in those paths and the mainnet runtime config is semantically unchanged, so there is no demonstrated causal link to this PR. The full dependency audit reports three high and five moderate toolchain or development advisories; this PR does not change dependencies or the lockfile, and the production-only audit is clean.

## Findings summary

- [P2] Keep the ADM explorer inside Tor
- [P2] Validate the emitted configuration, not the source file
- [P2] Reject missing and malformed endpoint lists

## Verification

- npm run wallets:data:check
- npm run lint
- npm run typecheck
- npm run test -- --run — 150 files and 1007 tests passed
- Focused ESLint over the newly added wallet and Vite files
- Production, testnet, and Tor PWA builds with compiled-chunk endpoint checks
- npm run electron:build:prepare
- npm run android:prebuild
- Negative mode checks for PWA tor-testnet, Electron testnet, and Android tor
- git diff --check
- npm audit --omit=dev — no production vulnerabilities

Recommendation: request changes until the three gate findings above are addressed.

vite-config/plugins/networkConfigPlugin.ts
adamant-al: Fixed. The Tor contract now requires onion hosts for `adm.explorer`, `adm.explorerTx`, and `adm.explorerAddress`, while the BTC, DASH, DOGE, and ETH explorer links keep the documented clearnet exemption.

Replacing the ADM Tor explorer URLs with mainnet values now fails the gate:

```
adm.explorer: Tor ADM explorer links must use onion hosts, found https://explorer.mainnet.example
adm.explorerAddress: Tor ADM explorer links must use onion hosts, found https://explorer.mainnet.example/address/${ID}
```

A unit test covers the rejection and keeps the third-party clearnet explorer links accepted, and the rule is documented in `CONTRIBUTING.md`.

vite-config/plugins/networkConfigPlugin.ts
adamant-al: Fixed. The gate no longer relies on module identity plus a fresh read of the source file. `generateBundle` now takes the rendered code of the bundled configuration module, and `findEmittedNetworkConfigViolations` compares the endpoint URLs in that code with the generated file. Extra or missing endpoints fail the build, and module code that the bundler does not expose fails the gate as well, instead of silently narrowing it to source validation.

Verified with your scenario: a `post` plugin that rewrites one Tor ADM node to `https://clown.adamant.im` in `transform`, after `load`. The Tor build fails:

```
Network configuration build gate rejected the tor bundle:
the emitted tor configuration contains https://clown.adamant.im, which the generated file does not declare
the emitted tor configuration does not contain http://37g5to2z6bdoeegun4hms2hvkfbdxh4rcon4e3p267wtfkh4ji2ns6id.onion
```

Unit tests cover matching code, tampered code, dropped endpoints, and unavailable module code. The mainnet, testnet, Tor, Electron, and Android builds still pass.

vite-config/plugins/networkConfigPlugin.ts
adamant-al: Fixed. `findNetworkConfigSchemaViolations` now runs before the URL and onion rules, and it fails closed:

- every coin and service of the mainnet configuration must be present
- every coin needs a non-empty `nodes.list`, and every service a non-empty `list`
- endpoint entries must be objects with a string `url` and an optional string `alt_ip`
- `nodes` and `services` sections that are not objects are rejected

`findNetworkConfigViolations('tor', {}, references)` now returns:

```
the network configuration has no coins
adm: missing coin configuration
btc: missing coin configuration
```

A string-valued section returns `btc.nodes: expected an object`, and an empty list returns `adm.nodes.list: expected a non-empty array of endpoints`. Unit tests cover these cases, and the gate list in `CONTRIBUTING.md` mentions the schema failure.

vite-config/plugins/networkConfigPlugin.ts
```diff
@@ -0,0 +1,483 @@
...
+      const configPath = getNetworkConfigPath(configsDir, variant)
+      const moduleIds: string[] = []
+      let emittedCode: string | null | undefined
+
+      for (const output of Object.values(bundle)) {
+        if (output.type !== 'chunk') continue
+
+        moduleIds.push(...output.moduleIds)
+
+        const emittedModule = output.modules[configPath]
```
dev-adamant-im: **[P2] Check the final output chunk after renderChunk**

`output.modules[configPath].code` is the rendered module code, not the final `output.code`. I ran a Tor Vite build with a review-only `renderChunk` hook that replaced an ADM onion endpoint with `https://clown.adamant.im`: the build passed, while the resulting JS chunks contained the clearnet URL and the module code still contained the original onion URL. A later chunk transform can therefore bypass the advertised emitted-bundle gate and break Tor endpoint isolation. Please validate the final runtime configuration as shipped, or otherwise verify that no later output transform changed it.

vite-config/plugins/networkConfigPlugin.ts
```diff
@@ -0,0 +1,483 @@
...
+export function findEmittedNetworkConfigViolations(
+  variant: NetworkConfigVariant,
+  config: NetworkConfig,
+  emittedCode: string | null
+): string[] {
+  if (emittedCode === null) {
+    return [`the bundler did not expose the emitted code of the ${variant} network configuration`]
+  }
+
+  const generatedUrls = new Set(collectConfigUrls(config))
```
dev-adamant-im: **[P2] Bind URLs to their config paths**

Comparing two sets of URL strings loses the fields they belong to. In a real Tor Vite build, a post-transform hook swapped `adm.nodes.list[0].url` (onion) with `btc.explorer` (intentionally clearnet); the build passed because the URL set was unchanged. Yet `findNetworkConfigViolations('tor', transformedConfig, references)` reports the clearnet ADM node. This can redirect ADM node requests to a clearnet explorer and violate the Tor privacy boundary. Please compare the structured path/value mapping of the emitted runtime config, rather than an unlabelled URL set.

vite-config/plugins/networkConfigPlugin.ts
```diff
@@ -0,0 +1,483 @@
...
+  reference: NetworkConfig
+): string[] {
+  if (!isPlainObject(config)) {
+    return ['the network configuration is not an object']
+  }
+
+  const coins = Object.keys(config)
+  const violations = coins.length === 0 ? ['the network configuration has no coins'] : []
+
+  for (const coin of new Set([...Object.keys(reference), ...coins])) {
```
dev-adamant-im: **[P2] Anchor the required runtime schema independently of mainnet.json**

`Object.keys(reference)` derives the required coin set from the generated mainnet file itself. If BTC disappears from both mainnet and Tor, `findNetworkConfigViolations` accepts both, although the BTC client directly reads `config.btc.nodes.list`. Removing `adm.nodes.healthCheck` or `adm.explorer` from Tor also passes because only endpoint lists are required, despite direct runtime reads of those fields. A pinned metadata regression can therefore make the drift check and this gate agree on an unusable configuration. Please validate a minimum client-required coin/service/field schema independently of the generated files.

vite-config/plugins/networkConfigPlugin.ts
```diff
@@ -0,0 +1,483 @@
...
+      }
+
+      const config = readNetworkConfig(configsDir, variant)
+      const violations = [
+        ...findBundledNetworkConfigViolations(variant, moduleIds, configsDir),
+        ...findNetworkConfigViolations(variant, config, {
+          mainnet: readNetworkConfig(configsDir, 'mainnet'),
+          testnet: readNetworkConfig(configsDir, 'testnet')
+        }),
+        ...(emittedCode === undefined
```
dev-adamant-im: **[P3] Fail closed when the selected module metadata is absent**

If `moduleIds` names the selected config but `output.modules` has no matching entry, `emittedCode` remains `undefined` and this branch adds no violation. A direct `generateBundle` probe with the Tor module ID, an empty `modules` map, and clearnet chunk code was accepted. I did not observe this metadata mismatch in normal Vite builds, but it contradicts the stated fail-closed behavior when emitted module code is unavailable. Please reject `undefined` here just as `null` is rejected.

vite-config/plugins/networkConfigPlugin.ts
adamant-al: Fixed. The selected generated config now resolves to a dedicated virtual module, and every HTTP(S) URL is marked with a path-bound assignment that survives later chunk transforms. `generateBundle` now inspects final `output.code`, compares the emitted path/value mapping against the generated config, and strips the temporary markers before write. Covered by the plugin spec and verified with `npm run build:tor`.

vite-config/plugins/networkConfigPlugin.ts
adamant-al: Fixed. The emitted-config check no longer compares unlabeled URL sets. It now compares the structured config path/value mapping recovered from the final chunk, so swaps such as `adm.nodes.list[0].url` and `btc.explorer` fail even when the raw URL set is unchanged. A focused unit case covers the swapped-path scenario.

vite-config/plugins/networkConfigPlugin.ts
adamant-al: Fixed. The schema gate no longer derives the required runtime contract from `mainnet.json`. The plugin now pins the client-required coins and services in code, requires explorer fields plus `nodes.healthCheck` and service `healthCheck` objects for them, and keeps the endpoint-list validation fail-closed. Added coverage showing the check still fails even when the reference configs omit the same fields.

vite-config/plugins/networkConfigPlugin.ts
adamant-al: Fixed as part of the final-chunk validation change. The build no longer skips emitted-config checks when module metadata is unavailable in `output.modules`; it validates the selected final chunk by module inclusion and fails closed when the marked runtime config cannot be recovered from that chunk.

adamantmm (approved):

PR #989, opened 2026-09-16
Fix: restore HTTP attachment CID hashing and notification i18n runtime handling

### LINKED ISSUE #988: [Bug] Fix HTTP attachment CID hashing and notification i18n runtime errors
# Summary

The web testnet client has two runtime regressions: file attachment selection fails on plain HTTP because CID hashing depends on `crypto.subtle`, and notification initialization can throw `this.i18n.t is not a function` when the app uses the composition-mode `vue-i18n` API shape.

# Steps to reproduce

1. Open `http://dev-testnet.adamant.im`
2. Log in and open any chat
3. Select a file to attach
4. Observe the console error `TypeError: Cannot read properties of undefined (reading 'digest')`
5. Open a build where notifications initialize from the current `Notifications` helper
6. Observe the console error `Uncaught TypeError: this.i18n.t is not a function`

# Expected behavior

File attachments should compute a CID and attach in both HTTP and HTTPS web contexts without weakening hashing guarantees.

Notifications should translate titles and messages without depending on one specific `vue-i18n` instance shape.

# Actual behavior

In insecure HTTP contexts where `crypto.subtle` is unavailable, attachment CID computation crashes before the file is attached.

Notifications can crash at runtime because the helper expects `this.i18n.t(...)`, but the injected translation API may only provide `$t` or `i18n.global.t`.

# Screenshots or videos

Console traces seen during reproduction:

- `TypeError: Cannot read properties of undefined (reading 'digest')`
- `Uncaught TypeError: this.i18n.t is not a function`

# Environment

- ADAMANT web client testnet builds
- macOS
- Browser context without `crypto.subtle` support because `isSecureContext === false`
- Reproduced on `http://dev-testnet.adamant.im`
- Compared against `https://dev-adamant-testnet.surge.sh`, where `crypto.subtle` is available

adamant-al:
## Description

- keep the browser-side CID hashing fast path in `src/lib/files/ipfs.ts` on `crypto.subtle.digest(...)` when Web Crypto is available and fall back to `@noble/hashes` plus explicit multihash digest construction only when it is not, so HTTP uploads work without regressing HTTPS responsiveness
- normalize notification translations in `src/lib/notifications.js` to use the injected `$t` function first, with compatible fallbacks for other `vue-i18n` shapes
- preserve the `Notifications` instance in `src/App.vue` so cleanup runs on unmount
- add focused regression tests for both computeCID hashing branches and the notification translation binding

## Related issue

Closes #988

## External links (optional)

- `http://dev-testnet.adamant.im`
- `https://dev-adamant-testnet.surge.sh`

Not included

## Breaking changes

No

## How to test

1. Open `http://dev-testnet.adamant.im`
2. Log in and open any chat
3. Attach a file and confirm it no longer fails with `Cannot read properties of undefined (reading 'digest')`
4. Confirm the app no longer throws `this.i18n.t is not a function` during notification initialization
5. Run `npx vitest run src/lib/files/ipfs.spec.ts src/lib/notifications.spec.ts`
6. Run `npm run typecheck`

## Notes for reviewers

- the attachment fix keeps SHA-256 as the hashing algorithm, preserves the native async Web Crypto path in secure contexts, and uses the pure-JS fallback only where `crypto.subtle` is unavailable
- the notification fix stays compatible with injected `$t`, `$i18n.t`, and `$i18n.global.t`

## Checklist

- [ ] Code is formatted
- [x] Tests added/updated (if needed)
- [ ] Documentation updated (if needed)
- [ ] Changes tested in all relevant environments


src/lib/files/ipfs.ts
```diff
@@ -9,7 +12,7 @@ export async function computeCID(fileOrBytes: File | Uint8Array) {
   const bytes =
     fileOrBytes instanceof File ? new Uint8Array(await fileOrBytes.arrayBuffer()) : fileOrBytes
 
-  const hash = await sha256.digest(bytes)
+  const hash = createDigest(SHA_256_MULTIHASH_CODE, sha256(bytes))
```
dev-adamant-im: Please keep the Web Crypto path when crypto.subtle is available and use @noble/hashes only as a fallback when it is not. computeCID handles encrypted attachments up to 250 MB; calling the pure-JS sha256(bytes) synchronously on every platform blocks the main thread for large files, including on HTTPS where the previous browser digest was asynchronous. This fixes the HTTP case but regresses responsiveness for secure contexts.

adamant-al: Restored the secure-context fast path here. `computeCID` now uses `crypto.subtle.digest("SHA-256", ...)` when Web Crypto is available and falls back to `@noble/hashes` only when it is not, so the HTTP fix stays intact without regressing responsiveness on HTTPS for large attachments. I also expanded the regression coverage to exercise both branches, and `npx vitest run src/lib/files/ipfs.spec.ts src/lib/notifications.spec.ts` plus `npm run typecheck` pass.

adamantmm (approved):

